PR #52, opened 2019-03-24
Test parsing definitions

JSAbrahams:
### Relevant issues
Definitions were untested.

### Summary
Now they are.
Uncovered bug where empty definitions incorrectly raised an error.
Uncovered a bug where expressions weren't always included in operations.
Uncovered bug where no error was raised if a statement in a block was not followed by a newline or dedent.

### Added Tests
Tests for multiple correct and incorrect definitions.

### Additional Context
...

### DIFF
--- a/tests/core/function.rs
+++ b/tests/core/function.rs
@@ -5,6 +5,7 @@ use mamba::lexer::tokenize;
 use mamba::parser::parse;
 
 #[test]
+#[ignore]
 fn core_function_definitions() {
     let source = valid_resource_content(&["function"], "definition.mamba");
     to_py!(source);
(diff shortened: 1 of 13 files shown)


tests/core/function.rs
```diff
@@ -5,6 +5,7 @@ use mamba::lexer::tokenize;
 use mamba::parser::parse;
 
 #[test]
+#[ignore]
```
JSAbrahams: Temporary, until the bug is fixed for parsing operations.

tests/parser/valid/function.rs
```diff
@@ -3,6 +3,7 @@ use mamba::lexer::tokenize;
 use mamba::parser::parse;
 
 #[test]
+#[ignore]
```
JSAbrahams: Temporary, until the bug is fixed for parsing operations.

PR #55, opened 2019-03-24
Cache kcov in travis

JSAbrahams:
### Relevant issues
Travis builds are taking quite long (5 minutes).

### Summary
Cache cargo and kcov, which should speed up the build process.

### Added Tests
Check build passes and that coverage report is generated.

### Additional Context
...


JSAbrahams:
Not that important right now, so closing for now.

PR #114, opened 2019-05-19
Remove with token in match and create With construct

### LINKED ISSUE #112: Change match syntax
### Current Issue
Currently `match` is a bit verbose, as writing `match <expression> with ...` is a bit cumbersome.
In particular the `with` keyword, which is used nowhere else and serves no real purpose here.

It might be better to do it like this: `<expression> match ...`, with `...` being the match arms.
This is similar to how Scala does it and to my eye a more elegant way to write match expressions.

Remove `with` keyword, and make `match` a postfix function.

JSAbrahams:
### Relevant issues
Fixes #112 

### Summary
Makes match more lightweight.
This token, however, can be used for a With construct (Similar to using resources in Python).


src/parser/statement.rs
```diff
@@ -42,3 +45,20 @@ pub fn parse_statement(it: &mut TPIterator) -> ParseResult {
         None => Err(CustomEOFErr { expected: "statement".to_string() })
     }
 }
+
+pub fn parse_with(it: &mut TPIterator) -> ParseResult {
+    let (st_line, st_pos) = start_pos(it);
+    check_next_is!(it, Token::With);
+    let resource = get_or_err!(it, parse_expression, "with resource");
+    let _as = if let Some(TokenPos { token: Token::As, .. }) = it.peek() {
```
JSAbrahams: double check that this is optional in Python

JSAbrahams: It is indeed optional.

Issue #19, opened 2019-03-15
Desugar calls to either python functions or descriptors


JSAbrahams:
### Description of the feature
Calls in the Mamba programming language can either be to a function with no arguments or can be a call to an attribute of an object.

In Python, `a.b` and `a.c()` are two different things.
In Mamba, We can call function `c` using `a.c`, `a.c()` and `a c`.
During the desugar stage we should either desugar based to a function call or descriptor access based on the type of `a` in this example.

### Description of potential solution
During the desugar stage there should be some sort of context which should allow us to get the type of an expression, and the accompanying functions.
With this list of functions, we can determine whether we wish to access a function or an attribute.

### Alternatives considered
It is also possible to check the type of everything again in the desugar stage.
However, this seems inefficient and would add a lot of unnecessary code, since this is already done in the type checking stage.

### Additional context
We should also think of a system of how to check python files directly.
Currently, the type checker will now only check the types, and the functions they contain, in Mamba files, and not Python files.

JSAbrahams:
Can use types defined in #65 and the state and context added in #71 .
Perhaps it might be better to annotate ASTNodes with types in the type checker, which will make it easier to check the type of identifiers to check whether they're a field/method, or if not, which fields/methods the class has of said identifier.

JSAbrahams:
#102 should simplify this somewhat, though we still need to check whether something is a function or a property (a function requires brackets in Python).

JSAbrahams:
Mamba has now been changed such that postfix notation is always interpreted as passing an argument to a function for the sake of readability.
To access properties we must always use dot notation.

PR #117, opened 2019-05-19
WIP Desugar default constructor and parents arguments

### LINKED ISSUE #53: Operations are not parsed fully
### Description of Bug
Operations are only parsed in situations where it starts with literal or `self`. 

### How to Reproduce
Take for instance `self.other + 10`
This is parsed as a function call, and it does not parse the operation fully.

### Expected behaviour
We would expect that any expression followed by a known operator should be parsed as an operation, with the correct order of operations.

### Additional context
...

### LINKED ISSUE #105: Desugar default constructors to init functions
### Current Issue
Default constructors are ignored.

Default constructors are constructors stored alongside the class name, i.e.: 
```mamba
class MyServer(mut self: DisconnectedMyServer, def ip_address: IPv4Address) isa Server
```

The above should be desugared such that we have an init function where these are assigned to self. So the above is roughly desugared to (self should be ignored):
```python
class MyServer(Server):
    ip_address = None

    __init__(self, ip_address):
        self.ip_address = ip_address
```

We desugar class nodes such that:
- The class gets the constructor arguments as properties (save for `self`).
- The class assign the constructor arguments to the properties of `self`.

The type checker should that that there is no other `init` function in the class.

JSAbrahams:
### Relevant issues
Fixes #105 
Fixes #53 
Addresses #19 

### Summary
- Move assignment outside `try` `except` for `handle` statements following an assign.
- Preserve type definitions, these become classes with empty methods in Python, which is the Python equivalent of an interface.
- Fix #53 by only parsing a single expression for postfix calls
- #19 is somewhat addressed by making property accesses with nothing following them property access. If we add brackets (with or without arguments), these do become functions. So the syntax behaves the same was as in Python.
  The intent was to someday allow users to call functions without brackets if these don't take arguments. So if we opt to still include this feature it will probably be included in V0.2


src/desugar/class.rs
```diff
@@ -0,0 +1,140 @@
...
+use std::ops::Deref;
+
+/// Desugar a class.
+///
+/// If a class has inline arguments (arguments next to class), then we create a
+/// constructor and assume that there is no constructor in the body of a class.
+/// This property should be ensured by the type checker.
+///
+/// We add arguments and calls to super for parents.
+pub fn desugar_class(node: &ASTNode, ctx: &Context, state: &State) -> Core {
```
JSAbrahams: This is the important part of this PR.
The application logic for desugaring both inline constructors and calls to parent is somewhat convoluted at the moment. I'm lookinto ways to simplify this.
I could probably split this into two major passes, one to check what has to be built, and a second one that actually builds the relevant nodes.
As it is now, these two stages are a bit intertwined (though this is also somewhat necessary as we want to desugar default and literal arguments as well).

JSAbrahams:
Probably best to only merge this after #123 so we can verify the output.

PR #123, opened 2019-05-22
Test AST's of test output

### LINKED ISSUE #122: Verify core output by verifying ast of Python output
### Current Issue
We don't systematically test that outputted Python files, so we have no way of checking they are semantically similar.
In most cases, we simply test that Mamba files can be parsed at the moment.

A simple way of doing this is by writing Python code and storing these in the test resources.
We can then compare these to the output of our pipeline.

We can then use a python library, or something along those lines, to convert both the Python resource and the outputted Python files.
This makes the test cases less brittle and cumbersome to write than doing string comparisons.
We can then easily, visually and quickly verify that outputted Python code is what we expect it to be.

JSAbrahams:
### Relevant issues
Fixes #122 

### Summary
Add tests to check that output ast is what we expect it to be.

- Fix not desugaring imports of file
- Remove multi-expressions from control flow, was more trouble than its worth
- Add raise construct to language, and `raise` token
- Fix not a desugaring overflow of `pow` operator, or `**` or `__pow__` in Python
- Always newline in control flow or function to prevent non-parsable Python code, such as broken if statements
- Fix not desugaring statement or expression contained in raises

- [x] Add Floordiv to core and add a test for floordiv when #115 is merged. 

### Added Tests
_List tests here once done_


JSAbrahams:
This PR is kind of on the large side, but I was a bit too lazy to split this up into smaller PR's.

PR #126, opened 2019-05-24
Refactor TP Iterator

### LINKED ISSUE #102: Use dot notation when accessing object fields or functions
### Current Issue
Currently, we use postfix notation both for passing arguments to functions, and when accessing properties and functions of objects.
This, however, might make code less readable.

Use dot notation for:
- Accessing functions of objects
- Accessing properties/fields of objects

We can still use postfix notation when passing arguments to functions, or when chaining function calls.

Just change the parsing rules.
As a bonus, this should make desugaring a lot easier.

### LINKED ISSUE #124: Use unreachable macro instead of unwrap
### Current issue
We use unwrap in multiple locations in the code which is discouraged in production code.

### description of feature
Add extra assertion checks for debug, and use `unimplemented!` macro combined with 'unwrap_or_else'.

In situations where we are **absolutely ** sure we may use `unreachable_checked`.

### LINKED ISSUE #125: Add functionality to TPIterator
### Current issue
We often use macros in the parse stage , and pass the iterator, which is ugly.
TPIterator is not much more than an alias at the moment 

### possible solution
Instead, add two (or more perhaps) functions to TP:
- `eat` to replace `check_next_is!'
- 'parse' to replace `get_or_err!` and `get_or_err_direct!`. This should take a function name (if possible) and error message as argument.

Design wise this also makes more sense since the iterator is stateful.

JSAbrahams:
### Relevant issues
fixes #125
fixes #124
fixes #102 

### Summary
Rather large refactor of the parsing stage.
More application logic has been moved to the `TPIterator` so this isn't scattered over the place.


JSAbrahams:
Turing out to be more ambitious than expected, especially due to Rust's borrow rules. So I will have to spend some extra time to make sure that the new design actually compiles is Rust.

PR #129, opened 2019-05-29
Switch from kcov to grcov

JSAbrahams:
### Relevant issues
codecov momentarily is unable to generate coverage reports.
It may be that there is an error in the kcov configuration, so I'm investigating whether this is the case.
Furthermore, grcov seems to be slightly more fleshed out than kcov.
At the very least, installing it locally is much easier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,5 +57,4 @@ script:
   - cargo test --verbose
 
   # Coverage report
-  - for file in target/debug/mamba-*[^\.d]; do mkdir -p "target/cov/$(basename $file)"; kcov --exclude-pattern=/.cargo,/usr/lib --verify "target/cov/$(basename $file)" "$file"; done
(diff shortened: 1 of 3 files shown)


JSAbrahams:
Thanks @casperboone for taking a look and identifying build issues on macos.

.travis.yml
```diff
@@ -57,4 +49,7 @@ script:
   - cargo test --verbose
 
   # Coverage report
-  - for file in target/debug/mamba-*[^\.d]; do mkdir -p "target/cov/$(basename $file)"; kcov --exclude-pattern=/.cargo,/usr/lib --verify "target/cov/$(basename $file)" "$file"; done
+  - export CARGO_INCREMENTAL=0
+  - export RUSTFLAGS="-Zprofile -Ccodegen-units=1 -Cinline-threshold=0 -Clink-dead-code -Coverflow-checks=off -Zno-landing-pads"
```
JSAbrahams: This does require nightly

PR #131, opened 2019-05-30
Transpile entire directories and create command line interface

### LINKED ISSUE #100: Transpile entire directories
### Current Issue
Currently we only transpile one file at a time.

When passed a directory, we transpile recursively every file in that directory.
This also requires that all this information be passed to the type checker before the desugaring stage so that the type checker can build a context with information regarding the entire directory.

Modify pipeline so:
- We iterate over each file, tokenizing and parsing each
- We pass the produced ast node array to the type checker. It will then:
   - Iterate over each, building a context
   - Iterate over each again, type checking them
- We pass the array of ast nodes and the produced context to the desugar stage, which will produce an array of core nodes. 
- For each core node, produce a transpiled python file.

We likely want to store a transpiled python project alongside its mamba counterpart.
Furthermore, it is also desirable to preserve the directory structure of the original project.
Therefore, when outputting python files:

- We store them in the same sub-directory relative to the new parent.

JSAbrahams:
### Relevant issues
fixes #100 

### Summary
Add options for transpiling an entire directory of files instead of a single file.

- Uses `input` as either input file or input directory
- Uses `output` as output directory (error if not directory)
- If no input, defaults to using `src` in current directory (error if this doesn't exist)
- If no output, defaults to using `target in current directory (creates one if this doesn't exist)

### Added Tests
Modified current tests to handle new command line interface.


tests/main.rs
```diff
@@ -12,24 +13,25 @@ mod desugar;
 mod lexer;
 mod output;
 mod parser;
+mod pipeline;
 
+// TODO make this test work
 #[test]
-fn command_line_class() -> Result<(), Box<std::error::Error>> {
+#[ignore]
+fn command_line_class_no_output() -> Result<(), Box<std::error::Error>> {
```
JSAbrahams: I have absolutely no idea why this fails. When I invoke the pipeline dirrectly it works just fine when I don't input. Somehow when I do this via the commandline however in tests this does not seem to work.
Perhaps I should invoke the executable directly and see if it works there as a sanity check.

Additionally, `println!` does not seem to print anything to console?
Currently, I use IntelliJ, and I know that its terminal is some kind of wrapper for the windows terminal (or I believe that I read it somewhere), meaning that sometimes the behaviour might differ from when an executable is called from the terminal. I experienced something similar when working on [Schaapi](https://github.com/cafejojo/schaapi), when using a logger, whose output was different within IntelliJ.

JSAbrahams:
It appears that the executable expects absolute paths as arguments instead of relative paths, which does not reflect the intended use-case. So this should be rectified.

tests/main.rs
JSAbrahams: Not really relevant after the application logic has been rewritten entirely.

PR #135, opened 2019-06-03
WIP Add doc strings and annotations

### LINKED ISSUE #119: Add doc strings
### Current Issue
In Python, we can add documentation using `"""`.
Mamba should have a similar construct.

Parse documentation and preserve this so Python code retains the docs written in Mamba.

Lex `"""` as doc token, and add this to the pipeline.

JSAbrahams:
### Relevant issues
Resolves #119 
Also adds annotations, which are useful.

### Summary
Add docs strings and annotations. These are either assigned to a variable, function, class or type.
While we could just handle them as statements, I think that it is better to couple these together with a node so that the type checker can check annotations. The parser can also ensure that we don't accidentally write documentation that is not linked to anything.

### DIFF
--- a/src/core/construct.rs
+++ b/src/core/construct.rs
@@ -1,103 +1,294 @@
 #[derive(Debug, PartialEq, Clone)]
 pub enum Core {
-    Module { id: String, imports: Vec<String>, body: Box<Core> },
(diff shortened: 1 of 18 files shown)


JSAbrahams:
We will redo this at a later date for a later release.

PR #136, opened 2019-06-04
Add osx to travis build

### LINKED ISSUE #134: Add OS X to CI build configuration
### Issue, or potential issue
We have no CI builds for mac OS.
We only have for Linux and Windows.
Currently, the project does build on Mac OS, but it is a better idea to use CI tools to check that this remains the case.
For a long time, for instance, the fact that the PYTHON value was not included in MacOS builds.

### Potential solutions
Add osx to Travis matrix.

JSAbrahams:
### Relevant issues
Resolves #134 

### Summary
Add `osx` to travis build conifg.
Now Travis is run on both linux and mac os.


JSAbrahams:
Travis is no longer being triggered at all, have to check why this is first.

PR #151, opened 2019-08-13
WIP Add creation of typing context (First pass)

### LINKED ISSUE #148: Build a context containing classes and their signatures
### Current Issue
Builds upon #147 

Do a preliminary analysis of all classes to get the names of all classes in a project and their accompanying functions.

**note** We do not have to check imports, though that should also be done at some point.

At a step to the type checker stage which precedes the current one, and which builds a context with mappings from classes to their accompanying functions and signatures.

Classes should store which interfaces they implement (Mamba as of now does not support inheritance, this is part of a discussion).
Function signatures should store function arguments, return type (if given), and errors the function may raise.

### LINKED ISSUE #154: Cannot have statements followed by a class in a script
### Description of Bug
The parser throws an error if we have a few statements followed by a class.

### How to Reproduce
Write a script with a few statements followed by a class

### Expected behaviour
We would expect the script to execute as usual
Namely, it should be possible to have a few statements followed by a class, i.e., write a script with some statements with for instance a helper class at the bottom.

JSAbrahams:
### Relevant issues
resolve #148 
fixes #154 

### Summary
Build a context with all the classes, fields, and functions.
This can be used for type-checking purposes when analysing expressions, such as checking that a function called on an expression indeed belongs to that type.

- Remove generics, will be re-added in future

Need to look further into namespaces.
At the moment, no import analysis is done either.

### Added Tests
- Building context of interfaces
- Building context of classes
- Check that self does not have a type
- Check that all arguments have a type when building context
- Check that if a function has no return type its return type is any (this is not the intended behaviour, so these tests will break once this behaviour has been changed, which we should do eventually)


tests/resources/valid/function/definition.mamba
```diff
@@ -3,6 +3,19 @@ def fun_a() =>
     if a and b then print c
     if c or d then print e
 
+class Something
+    def +(self, other: Int): Int => a + b
```
JSAbrahams: As these are inside Class something and override its methods, their return types should actually be `Something`, which can be verified by the type checker.

JSAbrahams:
Closing in favour of a new architecture.

PR #157, opened 2019-09-02
Build typing context

JSAbrahams:
### Relevant issues
We need a context with all the classes of an application to be able type check our application.

In future, we need to include classes of imported Python files as well.

### Summary
A typing context has all the relevant types, functions, and fields of the current application.

- Add `Position` and `EndPoint` structs to pipeline for easy passing of position without having to add a lot of dependencies where they shouldn't be. In effect, positions are abstracted away, which in future should make it easier to create more complex error messages.
- Add building of Context based on classes, functions, and fields of script. 
  - Types which may be inferred at a later date are wrapped in Optionals (in case they are not explicit yet).
  - An error is given when there was no type given when there should be one

### Added Tests
Rewrite current tests to use new `Position` struct


JSAbrahams:
Should add some extra tests before merging

PR #158, opened 2019-09-09
WIP Create basic system for generics

JSAbrahams:


src/type_checker/environment/type_name.rs
```diff
@@ -0,0 +1,93 @@
...
+                Ok(TypeName::Single {
+                    lit:      lit.clone(),
+                    generics: this_gens
+                        .iter()
+                        .map(|g| TypeName::try_from(g, generics, pos))
+                        .collect::<Result<_, _>>()?
+                }),
+                |generic_type_name| match generic_type_name {
+                    GenericTypeName::Single { lit, generics: other_gens } =>
+                        if other_gens.len() == this_gens.len() {
```
JSAbrahams: Actually, `other_gens` should just be empty, since it is never used.
For instance, we can have `List` as a generic parameter, without any generic arguments, and then it is given a generic parameter (in a correctly typed application), say:
```
def MyClass[A]
  def a: A[String]
```
Which would become
```
def MyClass[List]
  def a: List[String]
```
The error should instead occur if we try to pass List[String], which in the current implementation is seen as correct (with the above being incorrectly labeled a type error).
```
def MyClass[List[String]]
   def a: List[String][String] // This makes no sense
```

PR #181, opened 2019-11-24
Restrict access to private fields

### LINKED ISSUE #173: Outsiders can access private fields
### Description of Bug
Even if a function or field is marked as private, an outsider can still access these.

### How to Reproduce
```
class A
    def private my_field: Int <- 10


def a <- A()
# this should cause the type checker to error, but it doesn't
print a.my_field
```

### Expected behaviour
The type checker should error if we try to access a private field.

JSAbrahams:
### Relevant issues
Fixes #173 

See #188 for an identified issue.

### Summary
Private fields (both direct and defined in tuples) and functions cannot be accessed outside a class.

### Added Tests
- **Happy** Access field of a returned object of a function
- **Happy** Assign to a mutable field of a returned object of a function
- **Happy** Reassigning to mutable field in chain of calls to mutable objects
- **Ignored** **Sad** Assign to an mutable field of a returned object which is an immutable field of self returned by a function
- **Sad** Assign to an immutable field of a returned object which is a mutable field of self returned by a function
- **Sad** Assign to an immutable field of a returned object which is an immutable field of self returned by a function (see #188)
- **Sad** Assign to function call
- **Sad** Type definition with private field (interface cannot have private fields)
- **Sad** Type definition with private function (interface cannot have private function)
- **Sad** Accessing private field of object
- **Sad** Accessing private function of object
- **Sad** Accessing private field inside tuple of object
- **Sad** Reassigning to immutable field in chain of calls with an immutable object raises Type error


src/type_checker/infer/call.rs
```diff
@@ -83,6 +83,13 @@ fn property_call(
             let (instance, property_mutable, raises) = match &inner_instance.node {
                 Node::Id { lit } => {
                     let field = instance.field(&lit, &property.pos)?;
+                    if field.private && field.in_class != env.state.in_class {
```
JSAbrahams: Some code duplication due to sloppy implementation of recursion.
Should attempt to rewrite this file before merging if possible.

JSAbrahams:
Should add tests that test whether private tuples are indeed private, including their fields after #185 has been merged.

src/type_checker/infer/call.rs
```diff
@@ -77,78 +75,25 @@ fn property_call(
...
             let field = instance.field(&lit, &property.pos)?;
+            if field.private && field.in_class != env.state.in_class {
+                return Err(vec![TypeErr::new(&property.pos, &format!("{} is private", lit))]);
+            }
+
             let msg = format!("Cannot get type of field {}", field);
             let field_ty_name = &field.ty.ok_or_else(|| vec![TypeErr::new(&property.pos, &msg)])?;
-            Ok((InferType::from(&ctx.lookup(&field_ty_name, &property.pos)?), env.clone()))
+            let field_ty = InferType::from(&ctx.lookup(&field_ty_name, &property.pos)?);
+            Ok(((field.mutable, field_ty), env.clone()))
```
JSAbrahams: Not sure why this is not caught by the tests, but this should probably be a conjunction of the field being mutable and current mutable (`field.mutable && mutable`)

src/type_checker/context/field/generic.rs
```diff
@@ -42,13 +43,39 @@ impl TryFrom<&AST> for GenericField {
...
+    pub fn in_class(
+        self,
+        class: Option<&TypeName>,
+        type_def: bool,
+        pos: &Position
+    ) -> TypeResult<GenericField> {
+        if self.private && type_def {
+            Err(vec![TypeErr::new(
+                pos,
+                &format!("Field {} cannot be private: In an interface", self.name)
```
JSAbrahams: 'interface` -> `type definition`

src/type_checker/infer/call.rs
JSAbrahams: Realised this was slightly out of scope, but since this is so closely related I decided to fix it here and add tests for it here.

PR #182, opened 2019-11-24
Type check contents of f-Strings

### LINKED ISSUE #178: Expressions in fstrings are not type checked
### Description of Bug
Expressions in (f)strings are not type-checked

### How to Reproduce
```
def my_val: Int <- 100

# the below should throw a type error "my_vall is undefined"
print "my_val is {my_vall}"
```

### Expected behavior
Expressions in strings should be type-checked.

JSAbrahams:
### Relevant issues
Fixes #178 

### Summary
Add f-strings, which can have expressions contained within `{` and `}`.
I.e.:
```mamba
def my_name <- "Joël"
print "My name is {my_name}."
```
The expressions contained within the curly brackets are type-checked just as any other expression.
In addition to this, we verify that they are indeed expressions (and not statements) and that the resulting value has the `__str__` method.
This means that we are slightly more strict than Python, which would in the case of getting the string value of a value that does not have the `__str__` instead print its memory address:
```python
class A:
     my_field = 20


a = A()
/* prints <__main__.A object at 0x012OHD70> */
print(a)
```

In addition to this, strings which do not contain any expressions are treated and desugared as regular strings.
So:
`"my_string {my_value}"` becomes `f"my_string {my_value}"`
`"Hello World"` becomes/stays `"Hello World"`

### Added Tests
- **Happy** Expressions within f-Strings have the correct position(/offset) (which is useful for error printing
- **Happy** All primitives and collections can be contained within expressions within f-Strings
- **Happy** f-Strings containing expressions which have a type that implements `__str__` work as expected
- **Happy** `{` and `}` preceded by `\` are treated as regular characters
- **Happy** Empty expressions within strings encased in `{` and `}` are ignored
- **Sad** f-Strings containing expressions that contain an expression that doesn't implements `__str__` fails

### DIFF
--- a/src/type_checker/infer/literal.rs
+++ b/src/type_checker/infer/literal.rs
@@ -17,9 +17,9 @@ pub fn infer_literal(ast: &AST, env: &Environment, ctx: &Context) -> InferResult
             Node::ENum { .. } => ctx.lookup(&TypeName::from(concrete::ENUM_PRIMITIVE), &ast.pos),
             Node::Str { expressions, .. } => {
                 for expression in expressions {
-                    println!("{:?}", expression);
                     let (expr_ty, _) = infer(expression, env, ctx)?;
-                    expr_ty.expr_ty(&expression.pos)?;
+                    let expr_ty = expr_ty.expr_ty(&expression.pos)?;
+                    expr_ty.fun("__str__", &[], &expression.pos)?;
                 }
                 ctx.lookup(&TypeName::from(concrete::STRING_PRIMITIVE), &ast.pos)
             }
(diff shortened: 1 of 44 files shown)


src/type_checker/infer/literal.rs
```diff
@@ -15,7 +15,14 @@ pub fn infer_literal(ast: &AST, env: &Environment, ctx: &Context) -> InferResult
             Node::Real { .. } => ctx.lookup(&TypeName::from(concrete::FLOAT_PRIMITIVE), &ast.pos),
             Node::Int { .. } => ctx.lookup(&TypeName::from(concrete::INT_PRIMITIVE), &ast.pos),
             Node::ENum { .. } => ctx.lookup(&TypeName::from(concrete::ENUM_PRIMITIVE), &ast.pos),
-            Node::Str { .. } => ctx.lookup(&TypeName::from(concrete::STRING_PRIMITIVE), &ast.pos),
+            Node::Str { expressions, .. } => {
+                for expression in expressions {
+                    println!("{:?}", expression);
+                    let (expr_ty, _) = infer(expression, env, ctx)?;
+                    expr_ty.expr_ty(&expression.pos)?;
```
JSAbrahams: Should check that the expression has the __str__() method

PR #190, opened 2019-12-03
v0.2.1

JSAbrahams:
### Changelog
- Type check expressions within f-strings
- Fix parsing of tuples
  - Allow mutable fields within tuples
  - Fields within tuples are class fields if the tuple is a top-level declaration in a class
- Disallow access to private fields and functions from outside a class


JSAbrahams:
Double-check that release can be built before merging with master.

PR #195, opened 2020-01-04
Use constraints to type check language

### LINKED ISSUE #150: Check that a statement correctly mentiones that it raises an exception
### Current Issue
#148 and to a lesser extent #149 

In the language, any statement that may raise an exception must explicitly state so:
`<statement or expression> raises [<Errors>]`

Check that said expression or statement indeed does raise a said exception by looking at the signature of said function.

We also may have to look at the signatures of built-in functions, i.e. division can also throw an arithmetic error: `ZeroDivisionError`.
It would however also need to apply type sensitive flow here, i.e. if we beyond a doubt know that the denominator cannot be 0 we know for sure that this error will not be raised.

### LINKED ISSUE #166: Use contraint type checking instead of simple recursion
# Relevant issues

# Summary of current issue
Currently, we use simple recursion to type check our application.
However, this is rather limited, as the type inference stage can inference very little.

# Proposed change
Use constraints during type checking to allow for more advanced type checking.
Ideally, this should also make it much easier to implement for instance generics.
However, it will require us to closely keep track of the namespace of variables.

JSAbrahams:
### Relevant issues
Resolves #166 
Resolves #150 

### Summary
Use typing constraints for proper type inference.
This is all done in two stages:
1. Generation stage, where we generate constraints
2. Unification stage, where we solve these constraints like a set of equations. We also make use of substitution.

All this should also make it easier to type-check more complex language constructs in future.

We remove the old type inference logic, as well as any old (helper) functions which were used by the old type inference stage.

With regards to the `Node` class itself, we also did some cleanup by making mutable and the type properties of function arguments and variable definitions directly.
This means we don't constantly have to recursively check whether the identifier stored within these was an `IdType`.

We also make `TypeName` and `InferType` top-level in the type checker as these are universal constructs within the type checker.

### Added Tests

A few.

### Removed Tests

Some old tests have been removed due to them being to difficult to implement in one go.
We will gradually be re-adding language functionality, using these ignored tests as a roadmap.


JSAbrahams:
I know the build will fail, but I couldn't be bothered to fix it now.
I don't like to have `[ci skip]` in my git history, as this contains meta-information.

For now I will accept the failing build, so I don't just have a local copy of all my work.

PR #201, opened 2020-05-02
Integrate tuples into type checker

JSAbrahams:
### Relevant Issues
Implements #205 

### Summary
Keep track of tuple sizes, which can be statically determined, and give error if tuple sizes don't match during assignment.

- Mutability is not properly checked within tuples as of now.
- Parsing rules break down when we inherit from tuples. 
  We need to re-evaluate both the grammar of the language, and by extension the de-sugaring rules. 
  I think a lot of these rules can be generalized.
  Considering that I did struggle slightly at the start for creating grammar rules for classes, especially since I was not that familiar with how Python deals with classes and type aliases, re-evaluating the grammar here would be a good candidate for another PR.

### Added Tests
- *sad* assigning to tuple with wring number of arguments
- *sad* assign to size 2 tuple with single list


JSAbrahams:
Perhaps we should differentiate between tuples and list/sets/dictionaries (once we have dictionaries)

JSAbrahams:
Turning out to be a bit more difficult than anticipated, but the trace messages are proving very helpful.
It seems that the union stage is struggling with tuples, the constraints that are generated initially (and not during the unification stage) seem to be correct, but I have to double check this.
But I believe the error is in the unification stage currently, or there is some conceptual error in the generation stage.

src/check/env.rs
```diff
@@ -15,7 +15,8 @@ pub struct Environment {
     pub return_type: Option<Expected>,
     pub raises: Option<Expected>,
     pub class_type: Option<Expect>,
-    pub vars: HashMap<String, HashSet<(bool, Expected)>>
+    vars: HashMap<String, HashSet<(bool, Expected)>>,
```
JSAbrahams, line 18: The intersection and union methods should also handle these mappings.

src/check/constrain/constraint/expected.rs
```diff
@@ -27,37 +27,33 @@ impl Expected {
...
-            Node::Real { .. } => Type { name: NameUnion::from(FLOAT_PRIMITIVE) },
-            Node::Bool { .. } => Type { name: NameUnion::from(BOOL_PRIMITIVE) },
-            Node::Str { .. } => Type { name: NameUnion::from(STRING_PRIMITIVE) },
-            Node::Undefined => Nullable,
-            Node::Underscore => ExpressionAny,
-            _ => Expression { ast: ast.clone() }
-        };
-
-        Ok(Expected::new(&ast.pos, &expect))
+        Ok(Expected::new(&ast.pos, &Expect::from((ast, env))))
```
JSAbrahams, line 47: Double check that removed logic is still in generation stage.

JSAbrahams: Has been re-add, caused some tests to fail.

Issue #206, opened 2020-11-18
Use `fin` instead of `mut`


JSAbrahams:
### Current Issue
Mamba is intended to function much like Python, and be a scripting language.
Therefore, users familiar with Python will typically expect that when they define a variable they can directly edit it.

### High-level description of the feature
Instead of requiring that something be explicitly defined as mutable using the keyword 'mut', everything is mutable by default.
Instead, like in Java, we can explicitly define something as immutable by using the keyword 'fin`.
Unlike Java, however, when something is 'fin' (final), it is truly immutable, and its internal fields cannot be changed.
This also means that for methods which take `self` as argument, we can define self as `fin` so it will never change.

This is a bit of a difficult call to make, but I believe that considering the ecosystem of Python, Mamba should be more lenient.
However, this does mean that it is more likely that `self` will be mutable in a method, even when it shouldn't be.
Originally I wanted to prevent this, and took inspiration from Rust, but I think that Rust is perhaps more strict than what we're going for. 
In short:
- For short scripts, user's aren't burdened by needing the think about mutability.
- For larger systems, users can consider using `fin` to make the system more maintainable as it were.

### Description of potential implementation
-[ ] Drop 'mut' token
-[ ] Add 'fin' token
-[ ] Change behaviour in context/environment of type checker to deal with new tokens

JSAbrahams:
Already implemented

PR #207, opened 2020-11-18
Fin not mut

JSAbrahams:
### Relevant issues
#206 
Don't merge before #204 

### Summary
Change grammar slightly to further reduce complexity by removing `mut` keyword.
Everything is mutable by default now, and we use the `fin` keyword to specify that something is not mutable.

### Added Tests
Rewrote test resources to deal with change.


JSAbrahams:
We need to consider what happens when we.
Also, there seems to be a weird bug in the parsing of the function arguments, where if one still uses `mut` (or any other word) between `vararg`, and the rest of the argument, the parser doesn't fail but simply interprets this is another argument.
We would expect that the parser should fail because the function argument is malformed (there should be a comma between each argument) and not continue silently.

tests/output/valid/control_flow.rs
```diff
@@ -2,6 +2,7 @@ use crate::output::test_directory;
 
 #[test]
 fn for_ast_verify() -> Result<(), Vec<String>> {
+    loggerv::Logger::new().verbosity(3).module_path(false).init().unwrap();
```
JSAbrahams, line 5: Should not be merged

PR #209, opened 2021-02-15
Create rust.yml workflow file

JSAbrahams:
### Summary

Create workflow file to run tests on Ubuntu.
Replaces Travis and Appveyor.
Other tools have also been removed to streamline builds.

In future we can slowly re-add tools as we see fit.

Remove old badges from README.
In future, it would be nice to re-add these again.

TODO:
- [ ] Either check if Windows is option, and if so add it


JSAbrahams:
I'll look into windows in a future PR

JSAbrahams:
With windows, it seems that it cannot load the Python resources built into the type checker.
So I'm going to have to fix this soon, but not in this PR.

Issue #25, opened 2019-03-17
Lexer needs to be refactored


JSAbrahams:
### Current Issue

The lexer has become rather unwieldy and difficult to read.

### Proposed Solution

Now that tests have been added for the lexer, it will be easier to refactor the lexer. Therefore, it would be a good idea to refactor the lexer soon to make it more readable. The tests will help to make it easier to check for any potential regression.

JSAbrahams:
The lexer appears to be readable for now.
Might do another refactor in future, but for now it's fine.

PR #216, opened 2021-04-05
Update parser to conform to new grammar

JSAbrahams:
### Summary
Update parser so it is in accordance with https://github.com/JSAbrahams/mamba_doc/pull/25.
This should significantly simplify the logic of the parser as the grammar has been simplified and made more generic.

- The new grammar should also allow array indices to be defined, even if the type checker does not yet check these.
  That is something for a next milestone.
- If this is merged, I consider milestone v0.3.1 to have been pretty much reached, bar some extra testing of sad and happy paths to double check that we have the required features.

The diff is expected to be quite large due to the parser being quite large.
This will end up being quite an extensive re-write, even if the underlying logic is not that spectacular.


JSAbrahams:
No further pushes should be made until the underlying issues are fixed.
Re-starting development after a long break. Will get everything back to working order and fix this rather large change before working on more iterative changes and implementing smaller language features.

JSAbrahams:
Probably best to go over failing tests slowly and add language features.
For now we merge and iteratively start fixing language bugs, relevant to milestone 0.3.1

PR #218, opened 2022-01-02
Add a funny lines of code badge

JSAbrahams:


JSAbrahams:
Best  to double check the number actually means:
- Can we exclude python files?
- What is included and what isn't? (i.e. I'd say we _only_ want .rs files, no .lock file for instance)

JSAbrahams:
This seems unreliable, so for now we wont use it

PR #224, opened 2022-01-04
Allow tuples as class alias and restructure name

### LINKED ISSUE #223: Cannot use tuple as a class
### Description of Bug
When giving a class a tuple type, we get a type error where we expect an identifier.
Instead, this should be interpreted as a type.

### How to Reproduce
Compile:
```class MyTuple: (Int, Int, String)```
Get error:
```
[error | type] Expected identifier, was type
   1 | class MyTuple: (Int, Int, String)
                                       ^
```

### Expected behavior
Class should be a tuple type with the relevant types.

### PR DESCRIPTION
**Pull request by JSAbrahams**
### Relevant issues
Fixes #223

### CODE REVIEW

File: src/check/name/truename/generic.rs [hunk 1]
```diff
@@ -25,20 +25,25 @@ impl TryFrom<&AST> for TrueName {
     fn try_from(ast: &AST) -> TypeResult<TrueName> {
         match &ast.node {
             Node::Id { lit } => Ok(TrueName::from(&StringName::from(lit.as_str()))),
+            Node::Tuple { elements } => {
+                let elements = elements.iter().map(Name::try_from).collect::<Result<_, _>>()?;
+                Ok(TrueName::from(&NameVariant::Tuple(elements)))
+            },
             Node::QuestionOp { expr } => Ok(TrueName::try_from(expr)?.as_nullable()),
-            Node::Type { .. } => Ok(TrueName::from(&StringName::try_from(ast)?)),
+            Node::Type { id, .. } => Ok(TrueName::try_from(id)?),
```
JSAbrahams, line 33: What about generics here?

src/check/constrain/generate/class.rs
```diff
@@ -24,13 +24,13 @@ pub fn gen_class(
         Node::Class { body: Some(body), ty, .. } | Node::TypeDef { body: Some(body), ty, .. } =>
             match &body.node {
                 Node::Block { statements } =>
-                    constrain_class_body(statements, ty, env, ctx, constr),
+                    constrain_class_body(statements, ty, env, ctx, constr, &ast.pos),
```
JSAbrahams, line 27: Actually, this method probably shouldn't take a position.
Especially since each statement itself has a position, which could be used to create more local errors.
Now, if we have say, a very big class, then the error messages will refer to a large body of code and not be very helpful.

src/check/context/arg/generic.rs
```diff
@@ -109,21 +109,21 @@ impl TryFrom<&AST> for GenericFunctionArg {
...
                     ty: match ty {
-                        Some(ty) => Some(NameUnion::try_from(ty.deref())?),
+                        Some(ty) => Some(Name::try_from(ty.deref())?),
                         None if name.as_str() == SELF => None,
                         None =>
                             if let Some(default) = default {
                                 Some(match &default.deref().node {
                                     Node::Str { .. } =>
-                                        NameUnion::from(clss::python::STRING_PRIMITIVE),
+                                        Name::from(clss::python::STRING_PRIMITIVE),
```
JSAbrahams, line 118: Feels like the newlines here are actually unecessary

src/check/name/python.rs
```diff
@@ -2,23 +2,23 @@ use std::ops::Deref;
...
+impl From<&Expression> for Name {
     fn from(value: &Expression) -> Self {
         match value {
-            Expression::Name(_) => NameUnion::from(&TrueName::from(value)),
-            Expression::TupleLiteral(_) => NameUnion::from(&TrueName::from(value)),
+            Expression::Name(_) => Name::from(&TrueName::from(value)),
+            Expression::TupleLiteral(_) => Name::from(&TrueName::from(value)),
             Expression::Subscript(id, exprs) =>
                 if id.deref() == &Expression::Name(String::from("Union")) {
                     let names: Vec<TrueName> = exprs.iter().map(to_ty_name).collect();
```
JSAbrahams, line 16: These two lines may be combined

src/check/context/clss/generic.rs
```diff
@@ -185,7 +185,7 @@ impl TryFrom<&AST> for GenericClass {
 }
 
 fn get_fields_and_functions(
-    class: &StringName,
+    class: &TrueName,
     statements: &[AST],
     type_def: bool,
 ) -> Result<(HashSet<GenericField>, HashSet<GenericFunction>), Vec<TypeErr>> {
```
JSAbrahams, line 191: Shouldn't this be a TypeResult? It's shorter

JSAbrahams:
Please address clippy warnings first before merging.
Also it should fail if there are any clippy warnings though, so the build script should be fixed actually.

PR #225, opened 2022-01-04
WIP Also check None type in type checker

JSAbrahams:
### Summary
Before, none was ignored partially by the unification stage.
Now, none is just seen as any other type meaning that the same application logic is used to type check none types.


JSAbrahams:
Still issues where constraints appear to be improperly generated, such that the left side is not the superset/supertype of the right side even when the input is correctly typed (according to the language specification).

I.e.:
```def f(a): Int? => 10```

Is valid, as `10` is a `Int` and `Int?` is a supertype of `Int`, meaning that it should pass the type checking stage.
Need to fix how some constraints are generated before merging.
The same error appears also in simple arithmetic operations, not just functions, so not sure if this is an easy fix.

PR #231, opened 2022-01-06
Add grcov and codecov, clippy

JSAbrahams:
Also add clippy warnings to a pull request.
Ideally, pull requests should not be merged which have formatting issues either.

Tooling has up until this point been somewhat neglected.
We want to for now be up to date on the coverage, and ideally only merge pull requests which do not, or only marginally, decrease coverage.
The latter should be strongly motivated.
However, seeing as we're writing a transpiler, I'd argue that it is certainly possible to test a very large portion of the codebase.


JSAbrahams:
Some small fill formatting changes such that rustfmt should not complain

JSAbrahams:
Should note that the clippy action catches a lot more than when I run it locally.
Not quite sure what's happening there.

PR #234, opened 2022-01-07
Test suite overhaul and merge lexer and parser

### LINKED ISSUE #63: Overhaul test suite
### Issue, or potential issue
Currently, the test suite is still somewhat manageable, but as the codebase grows, this will become more of an issue.
There are still some slight inconsistencies, and some things are not as elegant as they should be, such as error reporting or test setup and cleanup.

A few identified issues:
- [x] The error messages generated by util are not descriptive because we don't do anything with the results, we immediately `unwrap()`. This Instead, we should actually do something when for instance a file is not found.
- [x] Function names in util are a bit long and unwieldy, we should consider changing these to make these more modular.
- [x] As we slowly move to the phase of testing outputted files based on input (see #17 and more importantly #32), we need to more systematically set up en clean up tests.
- [x] Tests should return a result (`Result<(), Box<std::error::Error>>`)

### Potential solutions
The [rust-lang-nursery](https://rust-lang-nursery.github.io/cli-wg/tutorial/testing.html) provides some information about testing applications.

### Additional Context
...

JSAbrahams:
### Relevant issues
Fixes #63 

- Tests return results, which are pretty printed to make any errors that show up more readable, thus hopefully speeding up development.
- Unit tests are moved to be alongside the source code instead of within the test suite. This is common practice within Rust.
- Temporary directories are properly deleted after tests, even after failure.


JSAbrahams:
Need to sleep on this, some unit tests borderline look like integration tests, in which case they should not be alongside the source.
I used the rule of, if it uses a test resource, it belongs in the test directory.
However, this is somewhat arbitrary I'd say.
There is not much difference between passing a string directly, and parsing a file first and then passing the string encoding the file contents.
So the grouping perhaps does not make much sense.

JSAbrahams:
I'd say that, since the lexer and parser are so tightly integrated, it makes sense to see them as unit tests.
As for the type checker and core, the distinction is perhaps less straightforward.
It would look like a unit test if we would manually constuct an AST tree, but this is far more work than its worth.
However, the functioning of type checker tests is tightly integrated with the success of the parser.
Perhaps then it is best to treat these as integration tests, though the structure is very similar to the parse unit tests so I'm a bit on the fence here.

JSAbrahams:
Merging the two actually now makes the distinction between unit and integration pretty clear now, check and core now are integration.
Output is more of a system test.
On a sidenote, core tests are integration as it does not run the type checker.

PR #239, opened 2022-05-07
Flip function access constraint generation

### LINKED ISSUE #228: Unification stage often fails when unifing accesses
### Description of Bug
Unification stage appears to fail when performing access:

```
class A
    def (a, b): (Int, Int) := (10, 100)


def a := A()
print a.a
print a.b
```
Gives the following error:
```
[error | type] Cannot infer type. Expected a String, was `a.a`.__str__(`a.a`)
  6 | print a.a
            ^^^
```

This happens also with primitives:
I.e.:
```
[error | type] Cannot infer type. Expected a String, was `self.required_field@0`.__str__(`self.required_field@0`)
  22 |     def fun_a(self: SomeState) => self.some_field := "my field is {self.required_field}"
                                                                          ^^^^^^^^^^^^^^^^^^^
  23 |     def fun_b(self) => print "this function is private: {self.private_field}!"
````

More in-depth description might follow.

JSAbrahams:
### Relevant issues
Resolves #228 

### Summary
Now, the thing the function return type is being compared to must be the super.
This makes sense, as a function may return a subset of the type its value is being assigned to,
but not the other way around.

- Streamline env.get_var method. Doesn't make
  sense to have function argument in return.

### Added Tests
- Unify access of a tuple arguments when tuple class attribute
- Tuple does not have type annotation when variable definition

### DIFF
--- a/src/check/constrain/generate/call.rs
+++ b/src/check/constrain/generate/call.rs
@@ -35,7 +35,7 @@ pub fn gen_call(
                     errors.push(TypeErr::new(&ast.pos, &msg))
                 }
 
-                if let Some((var, expecteds)) = env.get_var(var) {
+                if let Some(expecteds) = env.get_var(var) {
                     if expecteds.iter().any(|(is_mut, _)| !is_mut) {
                         let msg = format!("{} was declared final, cannot reassign", var);
                         errors.push(TypeErr::new(&ast.pos, &msg))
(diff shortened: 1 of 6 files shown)


JSAbrahams:
Well #228 has been resolved, but you can't actually annotate tuples in Python so I'll have to modify the desugar stage.

PR #241, opened 2022-05-07
v0.3.1 (Tuples)

JSAbrahams:


JSAbrahams:
Hm well nice that codecov has decided to work (need to check how it deals with develop).
However, guess the coverage has decreased more significantly than I expected, so perhaps I should write some more tests to make sure all the additional functionality works as intended.

JSAbrahams:
Ignore code coverage for now, I don't feel like testing every nook and cranny in this release.
Best to more extensively test features as they are built or reworked.
For instance, in this release I focused mostly on testing some functionality with regards to tuples to make sure this works as intended, to go beyond some of the end-to-end tests which don't cover all functionality.
Ideally we should in future have many more test scripts to test more "advanced" and niche language functionality, though.
I'd like to see every single language feature being tested in at least one end-to-end test eventually.

PR #256, opened 2022-05-11
Add index language construct

### LINKED ISSUE #251: Parse list indices
### Current Issue
Mentioned in #172 

Mamba cannot parse list indices.

Parse it somewhat similar to operation, if expression followed by `[` parse as list index.
What is within that operator is a range, which has a :
- Start (Optional (negative) Int)
- End (Optional (negative) Int)
- Step (Optional (negative) Int)

With the following structure:
```
range :=  [ [ '-' int] [ ( ':' ) | ':' [ '-' ] int ) ] [ [ - ] int ]
```

### LINKED ISSUE #252: Type Check Indices And Indexed Access
### Current Issue
Mentioned in #172 
Builds on #251 

Type check list indices:
- Check that arguments Int type
- Check that expression which is being indexed is one of:
  - List[Any]
  - Tuple[Any]
  - Dictionary[Any]

JSAbrahams:
### Relevant issues

Closes #251 
Closes #252

See #260 for discussion about functionality which could not fully be implemented in this PR.

### Summary

Implement indexes for collections.
An index may either be:
- An `Int`
- A `Slice`

Dictionaries ignored for now.
If they are implemented, then an index may be something other than a slice or integer.

### Added Tests

- Bunch of access with slice
- Access with an integer
- Functions which return a range
- Functions which return a slice
- Using a slice from a function to index an array
- Range and slice de-sugared properly


JSAbrahams:
Actually, slices and indexes should be different language constructs, since Python does treat these slightly differently.
This is especially complicated when we want functions to return either a `Slice` or a `Range`.
If they are the same language construct, then we will run into issues when we have a function returning either.
If Mamba has no way to differentiate the two, then both type checking and desugaring will be messy and produce applications with convoluted behavior.
Or, Mamba will say an application that appears well typed is not well typed, for instance.

JSAbrahams:
Probably would've been best to do this in little bits and pieces over multiple PR's (and to start with tests first), but I guess we'll do it like this.

spec/grammar.md
```diff
@@ -80,7 +84,7 @@ The grammar of the language in Extended Backus-Naur Form (EBNF).
     operation        ::= relation [ ( equality | instance-eq | binary-logic ) relation ]
     relation         ::= arithmetic [ comparison relation ]
     arithmetic       ::= term [ additive arithmetic ]
-    term             ::= inner-term [ multiclative term ]
+    term             ::= inner-term [ ( multiclative | range ) term ]
```
JSAbrahams, line 87: Add slice

src/check/constrain/generate/call.rs
```diff
@@ -41,13 +43,19 @@ pub fn gen_call(
                         errors.push(TypeErr::new(&ast.pos, &msg))
                     }
                 } else {
-                    let msg = format!("{} is undefined here.", var);
+                    let msg = format!("Variable '{}' is undefined in this scope.", var);
```
JSAbrahams, line 46: Remove "Variable"

src/check/constrain/generate/operation.rs
```diff
@@ -29,26 +30,23 @@ pub fn gen_op(
...
             generate(left, &env, ctx, &mut constr)
         }
-        Node::Range { from, to, step, .. } => {
-            let name = Name::from(clss::INT_PRIMITIVE);
-            let int_exp = &Expected::new(&from.pos, &Type { name });
-
-            constr.add("range from", &Expected::try_from((from, &env.var_mappings))?, int_exp);
-            constr.add("range to", &Expected::try_from((to, &env.var_mappings))?, int_exp);
-            let col = Expected::new(&ast.pos, &Collection { ty: Box::from(int_exp.clone()) });
-            constr.add("range collection", &col, &Expected::try_from((ast, &env.var_mappings))?);
```
JSAbrahams, line 39: Hm, if we treat the whole AST as a collection again for both slice and range it might fix our problem. Doesn't exactly feel elegant, perhaps in future we should differentiate between collection and iterable, though functionally they are identical in the current type checker I believe.

tests/resource/valid/access/index_via_function_check.py
```diff
@@ -0,0 +1,8 @@
+def f():
```
JSAbrahams, line 1: Make sure to update this file (but the test should fail anyway currently)

PR #261, opened 2022-05-13
WIP Add rustfmt to toolchain

### LINKED ISSUE #258: Add Rustfmt to CI pipeline
The build should fail if rustfmt complains.
This should also prevent big diffs in the future.
This should also prevent IDE specific peculiarities, such as InteliJ having an option where you can use either the internal linter or Rustfmt.

- [ ] Update Rustfmt config file.

JSAbrahams:
### Relevant issues

Closes #258

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -13,14 +13,18 @@ jobs:
     runs-on: "ubuntu-latest"
 
     steps:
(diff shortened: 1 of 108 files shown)


JSAbrahams:
Hm an action which reformats code is too aggressive I think.

PR #268, opened 2022-05-13
Treat Range as List[Int]

### LINKED ISSUE #260: Range is not properly seen also as a collection of integers
### Description of Bug

Following #256 , the type checker does not quite elegantly deal with the fact that a Range is also a collection of integers.
Though the converse is not true, of course.
I won't fix it in that PR quite yet.

### How to Reproduce

```
def f() -> Range => 0 .. 2 .. 1

for x in f() do
    print(x)
```
With the following error:
```
Error: [error | type] Unifying type: Expected a '{`x`}', was a 'Range'
 --> path/to/file:3:5
   3 | for x in f() do
           ^
   4 |     print(x)
```

Iterating directly over a range does appear to work, however:
```
for x in 0 .. 2 .. 1 do
    print(x)
```
So the type checker seems to make an exception there, though it shouldn't.
This is because currently it doesn't actual treat the above as a `Range` type at all.
It should deal with the above and that situation in an identical manner.

### Expected behavior

It should work.
Perhaps in future the type checker should instead check for iterables.
One can iterate over a range, but not a slice for instance.
The way collections have been implemented has been somewhat ad-hoc anyway.

JSAbrahams:
### Relevant issues

Closes #260

### Summary

Insert a `Range(List[Int])` class in built-in Python resources.
The type checker should detect that this is a type of collection, and then match the `List[Int]` with `List[Int]` when matching a `Collection { exp }` with `Type { .. }`.

Expand generics so that to make use of the fact that, for instance, Range is a type of List[Int] which is a type of Collection[Int].

### DIFF
--- a/src/check/constrain/unify/ty.rs
+++ b/src/check/constrain/unify/ty.rs
@@ -163,7 +163,7 @@ pub fn unify_type(
         (l_exp, r_exp) => match (l_exp, r_exp) {
             (Collection { ty }, Type { name }) | (Type { name }, Collection { ty }) => {
                 if let Some(col_ty) = name.collection_type(ctx)? {
-                    let expect = Expect::Type { name: col_ty.clone() };
+                    let expect = Expect::Type { name: col_ty };
                     constraints.push("collection type", ty, &Expected::new(&left.pos, &expect));
                     unify_link(constraints, ctx, total + 1)
                 } else {
--- a/src/check/constrain/unify/ty.rs
+++ b/src/check/constrain/unify/ty.rs
@@ -11,7 +11,7 @@ use crate::check::constrain::Unified;
 use crate::check::constrain::unify::expression::substitute::substitute;
 use crate::check::constrain::unify::link::unify_link;
 use crate::check::context::{Context, LookupClass};
-use crate::check::name::{IsSuperSet, Name};
+use crate::check::name::{CollectionType, IsSuperSet, Name};
 use crate::check::name::namevariant::NameVariant;
 use crate::check::result::{TypeErr, TypeResult};
 use crate::common::position::Position;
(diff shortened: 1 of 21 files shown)


src/check/name/stringname/mod.rs
```diff
@@ -35,8 +33,26 @@ impl Display for StringName {
     }
 }
 
+impl CollectionType for StringName {
+    fn collection_type(&self, ctx: &Context) -> Option<Name> {
```
JSAbrahams, line 37: Need to think about this method. What should it return?
Ideally we wan't to go beyond collections of just integers, but it should be anything.

src/check/constrain/unify/ty.rs
```diff
@@ -160,14 +160,26 @@ pub fn unify_type(
...
 
-        (l_exp, r_exp) => {
-            if l_exp.is_none() && r_exp.is_none() {
-                unify_link(constraints, ctx, total)
-            } else {
-                let msg = format!("Unifying type: Expected a '{}', was a '{}'", l_exp, r_exp);
-                Err(vec![TypeErr::new(&left.pos, &msg)])
+        (l_exp, r_exp) => match (l_exp, r_exp) {
+            (Collection { ty }, Type { name }) | (Type { name }, Collection { ty }) => {
+                if let Some(col_ty) = name.collection_type(ctx)? {
```
JSAbrahams, line 165: If we know how many placeholders types there are, then we can just give a placeholder Type as argument (`@0`, `@1`, etc.).
So environment needs to give as output how many placeholder types there are for the current set of constraints to the unification stage.
Then we can generate a placeholder type here, pass this as argument, and let the unification stage do its thing.

src/check/name/stringname/mod.rs
JSAbrahams: See comment on Environemnt and unifcation stage below.
It should just take a Name argument, and assume that we want a `Collection`.
This should work properly as `Collection` is a parent of `List` (which is what we have as palceholder type).

src/check/resource/std/collection.py
```diff
@@ -1,6 +1,14 @@
...
-class set(Generic[T]):
+class collection(Generic[T]):
+    def __init__(self): pass
+    def __iter__(self) -> collection_iter[T]: pass
+
+class collection_iter(Generic[T]):
+    def __init__(self): pass
+    def __next__(self) -> T: pass
+
+class set(Generic[T], collection[T]):
```
JSAbrahams, line 11: Looks like the type checker cannot deal with generics split over two parents.
At least like that.

Issue #110, opened 2019-05-16
How strict should we be about pure functions


JSAbrahams:
### Issue, or potential issue
Pure functions in Mamba are almost pure, but not quite.
Namely, while we disallow reading from global mutable variables, we do still allow writing to mutable variables outside the function, which is still a side-effect.

The question is, do we want to disallow this?
While this would make functions pure by definition, it would perhaps also decrease their usability.
My main concern is that while this feature would be nice, it is perhaps too cumbersome to use resulting in developers not using the feature at all.

I often see discussions about functional programming languages, and how they are safe, but that they are also often too restrictive, resulting in them not being used at all.
I think this discussion is somewhat similar to such discussions.
Do we wish to be pragmatic, and make functions "almost pure", or do we want to make them truly pure?

JSAbrahams:
Okay it should not have any side-effects, so it shouldn't write to anything.

PR #271, opened 2022-05-13
Flesh out pure function explanation in README

JSAbrahams:
Made small comment on this in #110 

Also specify how one may reassign within a pure function:
- Anything defined within the function body is fair game
- Any argument passed to the function may be reassigned to for convenience
- One may not read any field of a argument which is not final (`fin`)
- One may not call a function of a argument which is not pure (`pure`)
- One may not write to reassign or write to a property of a passed argument


JSAbrahams:
Need more explanation of what side-effects are before merging.

PR #272, opened 2022-05-13
Specify nullable string in code examples

JSAbrahams:


JSAbrahams:
Should also change 'true' into 'True' in first server example.

PR #275, opened 2022-05-15
Rename desugar to generate

### LINKED ISSUE #273: Add README's to sub-directories
Don't know how, but despite not working on it for well over a year I've managed to remember what most things do.
However, it'd still be nice to have some README's in the sub-directories which explain roughly what everything does.
Some solutions were not as trivial (for instance, `Name` required a few iterations to get working).
Therefore, having README's in a few key locations would help.

But generally we should stick to high-level directories so that they're easy to find.
But the idea is similar to #250 , we want key documentation to live alongside the (relevant) code instead of a separate location which is much harder to maintain and more likely to become outdated.
That feels way more rigid anyway.
Perhaps in the future we could somehow extract the README's from the codebase using an automated pipeline somewhere else?

JSAbrahams:
### Relevant issues

Closes #273 

### Summary

We also restructure the project such that we now have a (1) parse, (2) check, and (3) generate stage.
The parse stage has a (1.1) lex stage, which tokenizes input and (1.2) parse stage which parses input.
The generate stage has a (3.1) desugar stage, which converts input to an interal representation close to Python code and (3.2) a stage which then converts this to a String representing Python source code.

The lex stage is now also hidden behind the parse interface.
Parse takes a string and outputs an AST, as opposed to taking a vector of tokens.
This is now produced on the inside.

This is also part of an effort to have some spec live alongside the code instead of some separate poorly maintained repo somewhere.
Added README for:
- src (pipeline). This is implicitly also the readme for the `pipeline` dir.
- `generate`
- `check`
- `parse`

Credits to @casperboone for suggesting that `convert` be renamed to `generate` instead, which is more descriptive and more in line with standard compiler terminology.


tests/common.rs
```diff
@@ -47,7 +47,7 @@ pub fn resource_content_randomize(valid: bool, subdirs: &[&str], file: &str) ->
     }
 
     let source_path = tempdir_in(source_path.clone())
-        .expect(format!("Could not create temp dir: {}",
+        .expect(format!("Could not create desugar dir: {}",
```
JSAbrahams, line 50: Should say temp

README.md
```diff
@@ -305,12 +305,12 @@ We do type checking before desugaring to improve the quality of error messages.
 
 ### Desugar to Python
 
-Convert the AST to a simpler core language which looks similar to Python.
-This internal language is then again converted to a string which represents Python code.
-All erros generated here are either indicative of (1) a language construct which is not yet implemented (fully) or (2) an internal error.
+Convert the AST to a simpler desugar.core language which looks similar to Python. This internal language is then again
```
JSAbrahams, line 308: Sentence should start on new line

README.md
```diff
@@ -305,12 +305,12 @@ We do type checking before desugaring to improve the quality of error messages.
...
+converted to a string which represents Python code. All erros generated here are either indicative of (1) a language
```
JSAbrahams, line 309: Sentence should start on new line.

README.md
```diff
@@ -305,12 +305,12 @@ We do type checking before desugaring to improve the quality of error messages.
...
+construct which is not yet implemented (fully) or (2) an internal error.
 
-Note that in future, the type checker should annotate the tree such that each node has a type.
-This will allow us to add type hints to all Python code we output, and to perhaps also more easily desugar more complex language construts (such
+Note that in future, the type checker should annotate the tree such that each node has a type. This will allow us to add
```
JSAbrahams, line 312: Sentence should start on new line.

Cargo.lock
```diff
@@ -283,9 +285,9 @@ dependencies = [
 
 [[package]]
 name = "predicates-core"
-version = "1.0.0"
+version = "1.0.3"
```
JSAbrahams, line 288: Hm looks like cargo did some internal magic even though the `cargo.toml` was not updated.
Ideally this should be in a separate PR, so investigate what went wrong here.

JSAbrahams:
Would be cool to include the mamba logo in each README at the top, to make it a bit more colourful.

README.md
```diff
@@ -20,13 +20,9 @@
     <br/>
     <a href="https://github.com/JSAbrahams/mamba/milestones">
     <img src="https://img.shields.io/github/milestones/open/JSAbrahams/mamba?style=for-the-badge" alt="Active milestones"/>
-    <a/>
+    </a>
 </p>
 
-<h1 align="center">
```
JSAbrahams, line 26: This should be put back

src/README.md
```diff
@@ -0,0 +1,17 @@
+<p align="center">
+    <img src="../image/logo.svg" height="200" alt="Mamba logo"/>
+</p>
+
+# Structure
```
JSAbrahams, line 5: Where are the emoji's?

src/parse/README.md
```diff
@@ -0,0 +1,32 @@
+<p align="center">
+    <img src="../../image/logo.svg" height="200" alt="Mamba logo"/>
+</p>
+
+# Parse
+
+The parser has two internal packages:
+
+- `ast`, which is where `AST` and `Node` are stored.
+  - `AST` is container which stores a `Position` and a `Node`.
```
JSAbrahams, line 10: is a container

src/parse/README.md
```diff
@@ -0,0 +1,32 @@
+<p align="center">
+    <img src="../../image/logo.svg" height="200" alt="Mamba logo"/>
+</p>
+
+# Parse
```
JSAbrahams, line 5: Add errors sub-header

src/check/README.md
```diff
@@ -0,0 +1,11 @@
+<p align="center">
+    <img src="../../image/logo.svg" height="200" alt="Mamba logo"/>
+</p>
+
+# Check
```
JSAbrahams, line 5: Add constraint generation stage. 
Also specify that errors can happen at any stage.

src/check/README.md
```diff
@@ -0,0 +1,56 @@
...
+- Unification is performed when we encounter two types.
+
+## Type Names
+
+Names are at the core of the type checker, and are used to check whether a type is what we expect it to be.
+
+- A `Name` is the interface. It contains a set of `TrueName`s, as it represents a type union.
+- A `TrueName` contains two booleans: (1) Whether something is nullable and (2) whether something is mutable. It also
+  contains a `NameVariant`.
+- A `NameVariant` contains three variants:
```
JSAbrahams, line 53: can be one of three

src/desugar/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+# Desugar
+
+Converts `AST` to Python code.
+
+The input `AST` is converted to a special Python-like AST. This is then in turn converted to a String which should be a
+properly formatted Python file.
+
+We get an error if:
+- `AST` not what we expect, which is indicative of an implementation error within the type checker which should catch
```
casperboone, line 13:
```suggestion
- `AST` is not what we expect, which is indicative of an implementation error within the type checker which should catch
```

JSAbrahams:
Since we're changing things up anyway, shouldn't we just rename `desugar` to `convert`, to make clear that we're converting the Mamba AST to python code.
Internally this convert does do some desugaring, but it seems more logical going forward.
It wouldn't conflict with the other open PR, and we can just rename this one to make clear that this is the most significant change, and the README's are a bonus of this PR then.
(Truth be told I can't be bothered to split this up right now in two PR's)

casperboone:
> Since we're changing things up anyway, shouldn't we just rename `desugar` to `convert`, to make clear that we're converting the Mamba AST to python code.
> Internally this convert does do some desugaring, but it seems more logical going forward.

I agree that it should not be called a desugarer. If you’d like to stick to “standard” compiler pipeline terminology this would be the code generation step. So it may be an idea to call it code generate.

src/desugar/README.md
JSAbrahams: Thanks!

JSAbrahams:
> > Since we're changing things up anyway, shouldn't we just rename `desugar` to `convert`, to make clear that we're converting the Mamba AST to python code.
> > Internally this convert does do some desugaring, but it seems more logical going forward.
> 
> I agree that it should not be called a desugarer. If you’d like to stick to “standard” compiler pipeline terminology this would be the code generation step. So it may be an idea to call it code generate.

Hm good point `generate` would be even more be descriptive than `convert`, and bonus points for using "standard" terminology. (I'd been meaning to re-read up on this but life tends to get in the way.)

JSAbrahams:
Well I don't feel like editing README's forever this week.
But expanding on fixing errors within them will be relatively low-hanging fruit these coming days.
Also, making iterative changes to README's within smaller PR's will make it easier to focus more on writing quality as we go on.
It's not like we need to have perfect README's from day one.

tests/main.rs
```diff
@@ -10,7 +10,7 @@ use crate::common::resource_path;
 mod common;
 
 mod check;
-mod core;
+mod desugar;
```
JSAbrahams, line 13: generate?

tests/common.rs
```diff
@@ -47,7 +47,7 @@ pub fn resource_content_randomize(valid: bool, subdirs: &[&str], file: &str) ->
     }
 
     let source_path = tempdir_in(source_path.clone())
-        .expect(format!("Could not create temp dir: {}",
+        .expect(format!("Could not create generate dir: {}",
```
JSAbrahams, line 50: temp dir

PR #281, opened 2022-05-17
Generalize Collection Type Lookup

JSAbrahams:
### Summary

Generalize how one looks at a type:
- Look at whether it defines `__iter__`, which then returns the iterator type.
- We lookup the class in the Context.
- We look at whether it defines `__next__`, as any proper iterator class should.
- We return the return type of this `__next__` function.

This solution means that the collection lookup generalizes to all types which define an iterator.
For instance, we can now also index `String`s, which also have an iterator.

Also final round of tests before the next release.
Shouldn't be too much since we typically test as features are implemented, but looking at it with a fresh perspective can't hurt though.

### Added Tests

- Add extra test to verify that all in union must implement a method, in this case __bool__.
- All range and slice options as return of a function.
- Iterate over union of different types which are both iterable.
- *Sad* iterate over union of different types which are both iterable, but one of which does not implement method needed in loop body (minus op),
- Access types (in `definition`), which used to be accessed improperly
- E-numbers, which was already working but just an extra check
- Re-add match test, which tests match statements and also matching identifiers


src/generate/convert/class.rs
```diff
@@ -284,37 +320,90 @@ mod tests {
...
+            let statement = statements.first().unwrap();
+
+            assert_eq!(
+                *statement,
+                Core::FunDef {
+                    id: Box::new(Core::Id { lit: String::from("__init__") }),
+                    arg: vec![
+                        Core::FunArg {
+                            vararg: false,
+                            var: Box::from(Core::Id { lit: String::from("self") }),
```
JSAbrahams, line 376: If you make this into a macro you could shorten this probably

JSAbrahams:
Well, if we're so close to 84% coverage, maybe sneak in a few extra tests here.

JSAbrahams:
Re-added all ignored tests which previously failed, quite a lot of bugs which had been solved in the meantime.

Issue #188, opened 2019-12-01
We can circumvent mutability checking my reassigning to returned immutable variables


JSAbrahams:
### Description of Bug
Due to Python returning shallow copies from functions, we can reassign to fields of values even if the values themselves are immutable.

This behaviour was observed in #181, which contains an ignored test `assign_to_inner_non_mut3`.

### How to Reproduce
```mamba
class A
    def c: C

class C
    def my_class: D <- D()
    def my_field_accessor(self) -> D => self.my_class

class D
    def mut my_field: Int <- 10

def mut a <- A()
# should raise an error but doesn't
a.c.my_field_accessor().my_field <- 20
```

### Expected behaviour
We expect a type error because while `my_field` is mutable, the class in which it is contained, `C`, stored as `my_class` within `c`, is not.

### Additional context
It might be best to take a look at Rust's borrow system.
In Mamba, we could:
- Always return a deep copy of a field using the [deepcopy](https://docs.python.org/3/library/copy.html#copy.deepcopy) method. This would have to be clearly communicated with users. 
This is perhaps not the best alternative, as a lot of people coming from Python might run into unexpected behaviour if they often reassign to variables returned values of functions. 
  E.g. using getters directly: `my_object.get_my_field().inner_field <- 10`, will no longer have any effect as we reassign to a deep copy which is immediately discarded.
- Give a warning if trying to return a mutable field of self (perhaps this should be a lint), and give an error when trying to return an immutable field of self, meaning that it explicitly has to be copied using `deepcopy`.
  This would, however, result in a lot of unnecessary `deepcopy` invocations, which while useful in low-level languages, seems rather verbose for a more Python type language.
- Similar to above, except in the cases where we return a field of self, we preserve it's mutability when recursively checking calls.
  If we do not access a field of self, we take the variable to be mutable.
  This, I believe, most closely aligns with the philosophy of Mamba, since it reduces the verbosity, but still preserves the correctness of our type checker.

See Rust's [borrow checker](https://doc.rust-lang.org/1.8.0/book/references-and-borrowing.html), which addresses an issue similar to ours (though their solution is more verbose due to Rust's low level nature).

PR #294, opened 2022-05-19
Add new assign tokens and check nested calls

### LINKED ISSUE #276: Perhaps introduce new assign tokens
Seeing as the assign token `:=` is two characters wide anyway, should we introduce a few others?

- `+=` operator (add and assign)
- `-=` operator (subtract and assign)
- `*=` operator (multiply and assign)
- `/=` operator (divide and assign)
- `>>=` binary left shift by given amount and assign
- `<<=` binary left shift by given amount and assign

### LINKED ISSUE #277: Change inequality token
Using a slash is a bit, complicated.

Do `!=` instead of `/=`.
Forward slash means divide anyway, which we would likely want to reserve for #276 .

JSAbrahams:
### Relevant issues

Closes #277 .
Closes #276 .

### Summary

- Make idenitifer an enum, stepping stone toward fixing #188 .
- Made note of #296.
- Made note of #297.
- Fix name of some magic functions in checker.
- Create separate `NodeOp` for operation names.
- Create separate `CoreOp` and `CoreFunOp` for conversion of operation names.
- Change `/=` to `!=` token as described in #277 .
- Add all new assign tokens from #276 .

### Added Tests

- All new assign tokens.
- Assign to variables of an object.


src/check/constrain/generate/call.rs
```diff
@@ -27,7 +27,12 @@ pub fn gen_call(
     constr: &mut ConstrBuilder,
 ) -> Constrained {
     match &ast.node {
-        Node::Reassign { left, right } => {
+        Node::Reassign { left, right, op } => {
```
JSAbrahams, line 30: TODO: implement logic for type checker here

src/check/constrain/generate/call.rs
```diff
@@ -212,12 +217,17 @@ fn property_call(
             constr.add("call property", &instance, &access);
             Ok((constr.clone(), env.clone()))
         }
-        Node::Reassign { left, right } => {
+        Node::Reassign { left, right, op } => {
```
JSAbrahams, line 220: TODO: implement logic for type checker here

spec/grammar.md
```diff
@@ -128,12 +128,12 @@ The grammar of the language in Extended Backus-Naur Form (EBNF).
 
 ## Notes
 
-An `expression` is used in a situation where an expression is required.
-This allows the parser to short-circuit if something is definitiely not an expression where it should be.
-However we cannot always know in advance whether something is an expression, e.g. when it is a function call. 
-Those cases should be verified by the type checker.
-An `expr-or-stmt` may be used when it does not matter whether something is an expression or statement, such as the body of a loop.
+An `expression` is used in a situation where an expression is required. This allows the parser to short-circuit if
```
JSAbrahams, line 131: Okay need to double check what tool exactly is formatting these markdown files.
I'd prefer to have each sentence start on a new line.
And whatever tool it is, a config file where I can specify the above would be great.

src/generate/convert/mod.rs
```diff
@@ -42,9 +44,26 @@ pub fn convert_node(ast: &AST, imp: &mut Imports, state: &State) -> GenResult {
         },
 
         Node::VariableDef { .. } | Node::FunDef { .. } => convert_def(ast, imp, state)?,
-        Node::Reassign { left, right } => Core::Assign {
+        Node::Reassign { left, right, op } => Core::Assign {
             left: Box::from(convert_node(left, imp, state)?),
             right: Box::from(convert_node(right, imp, state)?),
+            op: String::from(if let Some(op) = op {
+                match op.deref() {
+                    Node::AddOp => "+=",
```
JSAbrahams, line 52: These strings should be an enum

src/generate/convert/mod.rs
```diff
@@ -84,6 +103,10 @@ pub fn convert_node(ast: &AST, imp: &mut Imports, state: &State) -> GenResult {
         Node::LeOp => Core::LeOp,
         Node::GeOp => Core::GeOp,
         Node::QuestionOp { .. } => convert_ty(ast, imp, state)?,
+        Node::BLShiftOp | Node::BRShiftOp => {
+            let msg = format!("Cannot have top level: {}", ast.node);
```
JSAbrahams, line 107: Illegal top-level constructs seem to occur more often.
Should create helper function which just generates this error instead of doing it separately each time here.

src/parse/call.rs
```diff
@@ -77,14 +105,37 @@ mod test {
     use crate::parse::{parse, parse_direct};
     use crate::parse::ast::{AST, Node};
 
+    #[test]
+    fn op_assign() {
```
JSAbrahams, line 109: Remainder should be tested in an end-to-end test ideally.
Separate test to see if the correct Python code is generated for this AST is misleading as it always goes through a type checker first.

src/generate/convert/mod.rs
```diff
@@ -383,7 +406,7 @@ pub fn convert_node(ast: &AST, imp: &mut Imports, state: &State) -> GenResult {
     let core = if let Some(assign_to) = assign_to {
         match core {
             Core::Block { .. } | Core::Return { .. } => core,
-            expr => Core::Assign { left: Box::from(assign_to), right: Box::from(expr) },
+            expr => Core::Assign { left: Box::from(assign_to), right: Box::from(expr), op: String::from("=") },
```
JSAbrahams, line 409: See comment on enum, use that

src/generate/convert/class.rs
```diff
@@ -129,6 +129,7 @@ fn extract_class(
                         property: arg.clone(),
                     }),
                     right: arg,
+                    op: String::from("="),
```
JSAbrahams, line 132: See comment on enum

spec/character.md
```diff
@@ -52,6 +52,18 @@ Keyword | Use
 `,`  | Separates arguments in collections or function or method signatures
 `_`  | Anonymous value
 
+## Reassignment Operations
+
+Keyword | Use
+---|---
+`+=`  | Add left side to right side and assign to left side
```
JSAbrahams, line 59: Lots of left and right here.
Consider using `variable` to refer to the left side, and `value` to the right:
`+=` then has the description `Add value to variable, and assign to variable`
Perhaps the worst offender here becomes more readable:
`*=` has description `Raise variable to value, and assign to variable`

JSAbrahams:
It's jumped back up to 86,07, which does somewhat support a hunch I've had where we didn't test access enough which has allowed a lot of weird behaviour to slip through.

Good thing is that it's systemic, so if we fix it then it should have a knock-on effect and fix a lot of other issues with the langauge.
Fixing this will in particular become more important as we start to test compilation of larger projects with many nested classes, which the current checker cannot handle (at all).
I guess this is my fault for only testing simple scripts and not setting up a more complex project.
Though arguably that would've been more effort than its worth since you have to be very careful about which language features you include.

src/check/constrain/generate/call.rs
```diff
@@ -38,27 +38,56 @@ pub fn gen_call(
...
-            constr.add(
-                "reassign",
-                &Expected::try_from((left, &env.var_mappings))?,
-                &Expected::try_from((right, &env.var_mappings))?,
-            );
-            let (mut constr, env) = generate(right, env, ctx, constr)?;
-            generate(left, &env, ctx, &mut constr)
+            if let Some(op) = op {
+                // Create temporary nodes and check that
+                let node = match op.as_ref() {
```
JSAbrahams, line 58: This can be a helper function which takes as argument `left` and `right` (and all the other necessary stuff). 
This also allows us to have a descriptive name like `convert_assign_and_check`, to highlight that we're "desugaring" the construct before type checking so we can re-use the checker logic.
The preservation of position should mean that error messages still point to the correct location in source as we do here.

src/parse/ast/mod.rs
```diff
@@ -69,6 +69,8 @@ pub enum Node {
     EqOp,
     LeOp,
     GeOp,
+    BLShiftOp,
```
JSAbrahams, line 72: Do we actually need all these `Op` `Node`s?
Can't this be a separate enum? (like `NodeOp`?)
Pretty sure these can't be top-level constructs anwyay in the generation stage, but need to check this.

src/parse/ast/node.rs
```diff
@@ -181,9 +183,10 @@ impl Node {
                 ty: Box::from(ty.map(mapping)),
                 args: args.iter().map(|a| a.map(mapping)).collect(),
             },
-            Node::Reassign { left, right } => Node::Reassign {
+            Node::Reassign { left, right, op } => Node::Reassign {
                 left: Box::from(left.map(mapping)),
                 right: Box::from(right.map(mapping)),
+                op, // op is untouched
```
JSAbrahams, line 189: If we use an special `NodeOp` enum, then we don't need to explain ourselves in a comment here!

spec/character.md
```diff
@@ -52,6 +52,18 @@ Keyword | Use
...
 `_`  | Anonymous value
 
+## Reassignment Operations
+
+Keyword | Use
+---|---
+`+=`  | Add value to variable and assign to variable
+`-=`  | Subtract value from variable and assign to variable
+`*=`  | Multiply value with variable and assign to variable
+`/=`  | Multiply value by variable and assign to variable
```
JSAbrahams, line 62: Divide by

spec/character.md
```diff
@@ -52,6 +52,18 @@ Keyword | Use
...
+`^=`  | Raise value by variable and assign to variable
```
JSAbrahams, line 63: Raise variable by value

src/check/constrain/generate/call.rs
```diff
@@ -285,8 +305,9 @@ fn property_call(
 /// A property call may not be a tuple, however.
 fn check_reassignable(ast: &AST) -> TypeResult<Identifier> {
     match &ast.node {
-        Node::PropertyCall { property, .. } => {
-            let identifier = check_reassignable(property)?;
+        Node::PropertyCall { instance, .. } => {
+            // We need logic here to check with nested variables if they may be assigned.
```
JSAbrahams, line 309: Logic might not need to be here, but certainly somewhere else.

JSAbrahams:
Currently working on identifier.
Should've been an enum from the start, similar to `Name`.
Once the new logic has been implemented, the checker should work properly for nested calls as it will check types recursively in such situations.

src/generate/ast/node.rs
```diff
@@ -89,3 +81,87 @@ pub enum Core {
...
+                CoreOp::Assign => "=",
+                CoreOp::AddAssign => "+=",
+                CoreOp::SubAssign => "-=",
+                CoreOp::MulAssign => "*=",
+                CoreOp::DivAssign => "/=",
+                CoreOp::PowAssign => "**=",
+                CoreOp::BLShiftAssign => "<<=",
+                CoreOp::BRShiftAssign => ">>=",
+
+                CoreOp::Ge => "__gt__",
```
JSAbrahams, line 149: Doesn't the type checker define these strings?

src/generate/ast/node.rs
```diff
@@ -89,3 +81,87 @@ pub enum Core {
...
+    Sub,
+    Mul,
+    Div,
+    Pow,
+    Mod,
+    FDiv,
+}
+
+impl CoreOp {
+    pub fn maybe_from(lit: &str) -> Option<CoreOp> {
```
JSAbrahams, line 112: check if `maybe_` prefix is idiomatic. Else, it is pretty obvious from the function returning `Option`.

tests/resource/valid/operation/assign_types.mamba
```diff
@@ -0,0 +1,13 @@
+def a := 10
+
+a += 2
+a -= 3
+a *= 6
+
+# a will now become a float
```
JSAbrahams, line 7: a was always a float, actually, if this is properly typed

JSAbrahams:
Should resolve final small issues and then merge.
The remaining tests which have been ignored should be noted in the issue, and we should fix that in the next PR.
This prevents everything being fixed in this one PR.

After that, we can start looking at how to generate python match statements (which I believe was introduced in Python v3.9, so we might need to think about how we deal with that in terms of parsing Python code, for which we use a dependency which goes up to v3.8 it seems).

src/generate/ast/node.rs
```diff
@@ -89,3 +82,87 @@ pub enum Core {
     With { resource: Box<Core>, expr: Box<Core> },
     WithAs { resource: Box<Core>, alias: Box<Core>, expr: Box<Core> },
 }
+
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+pub enum CoreOp {
```
JSAbrahams, line 87: This enum should be split in two:
- Operation which designate an operation within python (`Ge`, `Le`, `Pow`, etc.).
- Operations which designate names of functions which overload operations (AddAssign, SubAssign ,etc.).

tests/resource/valid/class/assign_types_double_nested.mamba
```diff
@@ -0,0 +1,19 @@
+class Y(def a: Float)
+
+class X
+    def y: Y
```
JSAbrahams, line 4: Should be a test as well to check that all class variables are actually signed to in the constructor.
If not, then the type should be nullable.

spec/grammar.md
JSAbrahams: It's Intellij, but looks like there [isn't](https://youtrack.jetbrains.com/issue/IDEA-257108) an option yet to disable this automatically.

Issue #297, opened 2022-05-23
Type checker does not recursively check assign


JSAbrahams:
### How to Reproduce

```
class Y(def a: Float)

class X
    def y: Y
    def init(self, a: Float) =>
         self.y := Y(a)

def x := X(10)

x.y.a := x.y.a + 2
x.y.a := x.y.a - 3
x.y.a := x.y.a * 6

x.y.a := x.y.a / 7
x.y.a := x.y.a ^ 2

x.y.a := x.y.a << 10
```

Gives an error that the type checker does not recognize `x.y.a`

### Expected behavior

It should treat `x.y.a` as a `Float` in the above example.

Likely need to add more generation logic to fully check the above expression.

PR #300, opened 2022-05-24
Check that all class vars assigned to in constructor

### LINKED ISSUE #298: Should verify that all top-level class variables are assigned to if not nullable

JSAbrahams:
### Relevant issues

Closes #298 .

### Summary

Type checker checks that all class variables are assigned to in the constructor.

Should not conflict with class-levle declarations (i.e. `MyClass(def x:Int)`) as this already conflicts with constructors.
This means that the type checker gives an error in such a situation, and we needn't deal with it.

Some other parser stuff:
- Make assign always top-level in constructor
  - Had to move around (and simplify, which is nice) some `check` logic to deal with the new way ASTs of assigns and property calls are handled. 
    Meaning, property calls are now always embedded within assigns, and not the other way around where you have a chain of property calls and an assign at the very end which doesn't make much sense.
- Remove `Node` `Init` and `_Self`. Could've been separate PR, but it was making the application logic here needlessly complicated so I just removed it in one go.

### Added Tests

- Checker allows non-assignment to nullable values.
- Checker does not allow non-assignment to non-nullable value (in constructor).
- Reassign is top level instead of property access in parser

TODO Will add tests for top-level tuples as well.
TODO check if this all solves #297 .

Generally speaking, this should be another stepping stone towards fixing #188 .


src/check/constrain/constraint/builder.rs
```diff
@@ -78,6 +78,11 @@ impl ConstrBuilder {
         self.constraints[self.level].1.push(constraint.clone())
     }
 
+    pub fn current_class(&self) -> Option<StringName> {
```
JSAbrahams, line 81: Actually, we should also check for parent classes, right?

JSAbrahams:
Once we properly deconstruct the `IdentiCall` in the generation stage, I believe that this all should work.
We need to properly map between expressions and the associated calls though for this to work.

JSAbrahams:
when running `unassigned_tuple_second_nullable`, we have the following constraints on this branche:
```
Constr: {left}  Int? == `self.z`
Constr: {left}  `self.z` == X.z
Constr: {left}  Int == `self.y`
Constr: {left}  `self.y` == X.y
Constr: {left}  `self` == X
Constr: {left}  `type tuple` == (Int, Int?)
Constr: {left}  (Int, Int?) == `(y@0, z@0)`
Constr: {left}  `self` == X
Constr: {left}  `self@0.y@0` == Int
Constr: {left}  Int == Int
Constr: {left}  `self@0.y@0` == `self@0`.y
```
And the following on develop:
```
Constr: {left}  Int? == `self.z`
Constr: {left}  `self.z` == X.z
Constr: {left}  Int == `self.y`
Constr: {left}  `self.y` == X.y
Constr: {left}  `self` == X
Constr: {left}  `type tuple` == (Int, Int?)
Constr: {left}  (Int, Int?) == `(y@0, z@0)`
Constr: {left}  `self` == X
Constr: {left}  `self`.y == Int
Constr: {left}  Int == Int
```

PR #302, opened 2022-05-24
Use match from python 3.10 in generated source

JSAbrahams:


JSAbrahams:
Decrease appears to be due to simplification of application logic. So, that's a good thing in this case.

PR #305, opened 2022-05-24
Remove Print Token and Node

JSAbrahams:
### Summary

No need for this application logic.
Leftover remnant from a very early version of the langauge where we wanted to be able to use postfix notation for the print funciton.
However, singling out the print function felt weird.
Python v3 also removed this notation, for many reasions (I could link the blog post here but it's a quick google search really).

In future, we should be able to actually add the print function to the context.
However, we need to be able to specify that the input should implement `__str__`.

### DIFF
--- a/spec/keyword.md
+++ b/spec/keyword.md
@@ -91,7 +91,6 @@ Keyword | Use
 
 Keyword | Use
 ---|---
-`print`   | Print value of an expression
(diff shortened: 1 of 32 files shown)


JSAbrahams:
Drop in coverage is due to simplification of application logic, so here it's a good thing.

Issue #279, opened 2022-05-16
Type checker should annotate the AST with types


JSAbrahams:
### Current Issue

The type checker does not annotate the AST with types.
This makes more "advanced" language features, such as automatically returning the last expression (without `return`), or treating a match as an expression more difficult.

A few issues which would benefit:
- #278 wrapping literals. We only need to wrap them if we perform a function call. It would be redundant to wrap them in every location.

This also saves a lot of types which we can use to annotate our Python output.
Throwing this away is a shame.

JSAbrahams:
In the unification stage, we have the position of the asts.
One option is to return a set of positions and types at that position.
Then, for each position, we can recursively go through the AST and as soon as the position is found we annotate that `ASTTy` with the type.

JSAbrahams:
I doubt this feature will really help with #278 , but still nice to have of course.

PR #306, opened 2022-05-25
Streamline interfaces

JSAbrahams:
### Relevant issues

Working towards #279 

### Summary

- Add ASTTy class, which in future should represent an annotated type tree.
- Interfaces no longer take source and path as input (expect parse, which takes source as input of course).
  Instead, the user can opt to add these to the errors if they so choose.

### DIFF
--- a/src/check/ast.rs
+++ b/src/check/ast.rs
@@ -1,0 +1,30 @@
+use std::ops::Deref;
+
+use crate::check::name::Name;
+use crate::common::position::Position;
+use crate::parse::ast::{AST, Node};
+
+#[derive(PartialEq, Eq, Hash, Clone, Debug)]
+pub struct ASTTy {
+    pub pos: Position,
+    pub node: Node,
+    pub ty: Option<Name>,
+}
+
+impl From<&AST> for ASTTy {
+    fn from(ast: &AST) -> Self {
+        ASTTy { pos: ast.pos.clone(), node: ast.node.clone(), ty: None }
+    }
+}
+
+impl From<&Box<AST>> for ASTTy {
+    fn from(ast: &Box<AST>) -> Self {
+        ASTTy::from(ast.deref())
+    }
+}
+
+impl ASTTy {
+    pub fn to_ty(self, ty: &Name) -> ASTTy {
+        ASTTy { ty: Some(ty.clone()), ..self }
+    }
+}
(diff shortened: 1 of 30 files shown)


src/check/context/generic.rs
```diff
@@ -1,50 +1,34 @@
...
                         Node::Class { .. } | Node::TypeDef { .. } | Node::TypeAlias { .. } => {
-                            let generic_type: Result<_, Vec<TypeErr>> =
-                                GenericClass::try_from(module).map_err(|errs| {
-                                    errs.into_iter()
-                                        .map(|e| e.into_with_source(source, path))
-                                        .collect()
-                                });
-                            types.insert(generic_type?);
+                            let generic_type = GenericClass::try_from(module)?;
+                            types.insert(generic_type);
```
JSAbrahams, line 24: This can be one-lined

src/generate/mod.rs
```diff
@@ -59,26 +60,7 @@ pub type DesugarInput = (AST, Option<String>, Option<PathBuf>);
...
-        Ok(oks
-            .iter()
-            .map(|(res, src, path)| (res.as_ref().unwrap().clone(), src.clone(), path.clone()))
-            .collect())
-    } else {
-        Err(errs.iter().map(|(res, ..)| res.as_ref().unwrap_err().clone()).collect())
-    }
+pub fn gen(input: &ASTTy) -> GenResult {
+    let ast = AST::new(&input.pos, input.node.clone());
+    convert_node(&ast, &mut Imports::new(), &State::new())
```
JSAbrahams, line 65: Change signature of `convert_node` and all inner nodes to make use of `ASTTy`.
This is not a big change, and allows us to also use the type in annotations

src/check/ast.rs
```diff
@@ -0,0 +1,30 @@
+use std::ops::Deref;
+
+use crate::check::name::Name;
+use crate::common::position::Position;
+use crate::parse::ast::{AST, Node};
+
+#[derive(PartialEq, Eq, Hash, Clone, Debug)]
+pub struct ASTTy {
+    pub pos: Position,
+    pub node: Node,
```
JSAbrahams, line 10: Need new `Node` variant which has `ASTTy` children instead of `AST` children.
Going to be a lot of work, probably for a next PR.
This PR only changes the interface for now.

PR #309, opened 2022-05-26
Streamline import node

JSAbrahams:
### Summary

- Get rid of FromImport
- Get rid of ImportAs


JSAbrahams:
Decrease in coverage because we have removed some logic which was mostly covered.

PR #310, opened 2022-05-27
Add optional to Name in generate

JSAbrahams:
### Summary

Always print out generics in alphabetical order.
This ensures that the output is easy to predict, we are not dependent on the whims of the iterator over the set, possibly preventing flaky tests.

Make annotating output the default in tests as well. 
As we start to implement new features and debug the checking stage we should then automatically see old tests failing  as well.

Should create issues for all unexpected/wrong behavior before merging this PR.
At the very least we can now more clearly see what the checker is doing on the hood, which is great!

### Added Tests

- Check that generics are printed in alphabetical order in source. Prevents flaky tests as well.
- Make annotating output default for all tests, with option to turn this off via the `Arguments` struct.

### DIFF
--- a/src/generate/name.rs
+++ b/src/generate/name.rs
@@ -71,38 +71,174 @@ impl ToPy for StringName {
 
 /// Produce type with alphabetized generics, ensuring that for any two equal sets of generics the
 /// order in which they are given is always the same.
+///
+/// - Ignores capitalization of generics.
+/// - Types with no literal are at the front.
 fn core_type(lit: &str, generics: &[Core]) -> Core {
-    let names: BTreeMap<String, Core> = generics.iter().map(|core| match core {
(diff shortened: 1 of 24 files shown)


JSAbrahams:
Only unmark as draft _after_ having created issues for all unexpected behaviour. 
This is so they aren't forgotten.

src/generate/name.rs
```diff
@@ -56,7 +64,45 @@ impl ToPy for NameVariant {
...
+    Core::Type { lit: String::from(lit), generics }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::generate::ast::node::Core;
+    use crate::generate::name::core_type;
+
+    #[test]
+    fn alphabetize_ids() {
```
JSAbrahams, line 94: Add extra test for more complex nested structures (type in type)

PR #323, opened 2022-05-27
Use new super calls in generated

### LINKED ISSUE #314: Use incorrect initialisation of super class in generated code
### Description of Bug

The manner in which we generate calls to parents is incorrect it seems.

### How to Reproduce

Mamba source
```
class MyType(a: String)
class MyType2(b: String)

class MyClass1: MyType("asdf"), MyType("qwerty")
    def other: Int
```

Is converted to:
```python3
class MyType:
    def __init__(self, a: str):
        self.a = a

class MyType2:
    def __init__(self, b: str):
        self.b = b

class MyClass1(MyType, MyType2):
    other: int = None

    def __init__(self):
        super(MyType, self).__init__("asdf")
        super(MyType2, self).__init__("qwerty")
```

### Expected behavior

But super calls should be as follows:
```python3
        MyType.__init__(self, "asdf")
        MyType2.__init__(self, "qwerty")
```

### LINKED ISSUE #320: Generate stage creates constructor additional construtors with calls to super even if one already exists

JSAbrahams:
### Relevant issues

- Closes #314 
- Closes #320 

- Made note of #324


JSAbrahams:
Ah note to self: Branching from an unmerged branch, and then squash merging said branch, leads to a bit of an ugly commit history here.

PR #325, opened 2022-05-28
Remove File Node

JSAbrahams:
### Summary

- Move logic to parse statements, everything is now a block.
  This means that imports may also be nested, though the context does not yet deal with this.
- Add EOF token
- Treat import as statement.
- Make Position and CaretPos implement copy. 
  Then we don't have to constantly clone these simple constructs.
- Improve some error messages.


JSAbrahams:
Decrease in coverage is largely due to simplification of logic.

PR #328, opened 2022-05-28
Generate new type alias

### LINKED ISSUE #324: Cannot create type alias with primitives
### Description of Bug

Cannot create type alias with primitives.

### How to Reproduce

```
class MyGeneric: String
```

Produces the following python code:
```
MyGeneric = str
```
Which python does not like, it then expect a `:`, not a `=`.
This does work find though for other aliases.

### Expected behavior

It should create a valid type alias.

### Additional context

https://docs.python.org/3/library/typing.html#newtype

For instance, we would then have:
```
from typing import NewType

MyGeneric = NewType('MyGeneric', str)
```

This is a more safe way to generate type aliases.

JSAbrahams:
### Relevant issues

- Closes #324 
- Made note of #329 

### Summary

Aside from above issue, we now:
- Always initialize the super class with self.
- Preserve generics from parent.

### Added Tests

- Simple type alias

### DIFF
--- a/src/generate/ast/mod.rs
+++ b/src/generate/ast/mod.rs
@@ -47,9 +47,17 @@ fn to_py(core: &Core, ind: usize) -> String {
     match core {
         Core::Import { from, import, alias } => format!(
             "{}import {}{}",
-            if let Some(from) = from { format!("from {} ", to_py(from, ind)) } else { String::from("") },
+            if let Some(from) = from {
+                format!("from {} ", to_py(from, ind))
+            } else {
+                String::from("")
+            },
             comma_delimited(import, ind),
-            if !alias.is_empty() { format!(" as {}", comma_delimited(alias, ind)) } else { String::from("") }
+            if !alias.is_empty() {
+                format!(" as {}", comma_delimited(alias, ind))
+            } else {
+                String::from("")
+            }
         ),
         Core::Id { lit } => lit.clone(),
         Core::Type { lit, generics } => {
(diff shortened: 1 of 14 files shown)


tests/resource/valid/operation/type_alias_primitive_check.py
```diff
@@ -0,0 +1,4 @@
+MyType = newType('MyType', str)
```
JSAbrahams, line 1: `NewType`, also missing import

src/generate/ast/mod.rs
```diff
@@ -253,14 +261,14 @@ fn to_py(core: &Core, ind: usize) -> String {
         Core::Break => String::from("break"),
 
         Core::ClassDef { name, parent_names, body } => format!(
-            "class {}{}:\n{}\n",
+            "class {}{}:{}\n",
```
JSAbrahams, line 264: Add space

src/generate/convert/class.rs
```diff
@@ -1,13 +1,22 @@
...
-use crate::check::context::{arg, function};
-use crate::generate::ast::node::{Core, CoreOp};
-use crate::generate::convert::common::convert_vec;
-use crate::generate::convert::convert_node;
-use crate::generate::convert::state::{Imports, State};
-use crate::generate::result::{GenResult, UnimplementedErr};
+use std::{collections::HashMap, ops::Deref};
+use itertools::Itertools;
+
+use crate::{
```
JSAbrahams, line 4: Are we sure that this is what rustfmt config specifies?

PR #339, opened 2022-05-31
Fix check access of nested fields of object

### LINKED ISSUE #297: Type checker does not recursively check assign

JSAbrahams:
### Relevant issues

- Closes #297 

### Summary

Fix #297 by recursively for each inner call also generating a constraint.

#### An example

`x.y.z` is turned into:
- `'x'.y == 'x.y'`
- `'x.y'.z == 'x.y.z'`

Then in the unifcation stage:
- `'x'.y` is a direct access so one can take the class of `x`, look at field `y`, and determine the of expression `'x.y'`.
- One now knows what `'x.y'` is in `'x.y'.z`, so we lookup `z` in whatever type `'x.y'` is to know the type of `x.y.z`.

### Added Tests
- Operations on nested access
- Print of nested access

### DIFF
--- a/src/check/constrain/constraint/expected.rs
+++ b/src/check/constrain/constraint/expected.rs
@@ -17,7 +17,6 @@ use crate::check::name::string_name::StringName;
 use crate::check::result::{TypeErr, TypeResult};
 use crate::common::delimit::comma_delm;
 use crate::common::position::Position;
-use crate::common::result::an_or_a;
 use crate::parse::ast::{AST, Node};
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
@@ -128,7 +127,7 @@ impl Display for Expect {
                 Access { entity, name } => format!("{}.{}", entity.expect, name.expect),
                 Function { name, args } => format!("{}({})", name, comma_delm(args)),
                 Field { name } => name.clone(),
-                Type { name: ty } => format!("{}'{}'", an_or_a(ty), ty),
+                Type { name: ty } => format!("'{}'", ty),
             }
         )
     }
--- a/src/check/constrain/constraint/expected.rs
+++ b/src/check/constrain/constraint/expected.rs
@@ -127,7 +127,7 @@ impl Display for Expect {
                 Access { entity, name } => format!("{}.{}", entity.expect, name.expect),
                 Function { name, args } => format!("{}({})", name, comma_delm(args)),
                 Field { name } => name.clone(),
-                Type { name: ty } => format!("'{}'", ty),
+                Type { name: ty } => format!("{}", ty),
             }
         )
     }
--- a/src/check/constrain/constraint/expected.rs
+++ b/src/check/constrain/constraint/expected.rs
@@ -17,6 +17,7 @@ use crate::check::name::string_name::StringName;
 use crate::check::result::{TypeErr, TypeResult};
 use crate::common::delimit::comma_delm;
 use crate::common::position::Position;
+use crate::common::result::an_or_a;
 use crate::parse::ast::{AST, Node};
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
@@ -127,7 +128,7 @@ impl Display for Expect {
                 Access { entity, name } => format!("{}.{}", entity.expect, name.expect),
                 Function { name, args } => format!("{}({})", name, comma_delm(args)),
                 Field { name } => name.clone(),
-                Type { name: ty } => format!("{}", ty),
+                Type { name: ty } => format!("{}'{}'", an_or_a(ty), ty),
             }
         )
     }
--- a/src/check/constrain/constraint/iterator.rs
+++ b/src/check/constrain/constraint/iterator.rs
@@ -47,7 +47,7 @@ impl Constraints {
     /// Only used during unification stage.
     pub fn push(&mut self, msg: &str, left: &Expected, right: &Expected) {
         let constraint = Constraint::new(msg, left, right);
-        trace!("{:width$}[gen {}] {}", "", msg, constraint, width = 27);
+        trace!("{:width$}[gen {}] {}", "", msg, constraint, width = 2);
         self.constraints.push_front(constraint)
     }
 
--- a/src/check/constrain/constraint/iterator.rs
+++ b/src/check/constrain/constraint/iterator.rs
@@ -34,7 +34,7 @@ impl Constraints {
     pub fn push_ty(&mut self, pos: Position, name: &Name) {
         let name = self.finished.get(&pos).map_or(name.clone(), |s_name| s_name.union(name));
         if self.finished.insert(pos, name.clone()).is_none() {
-            trace!("{:width$} type: {}", pos, name, width = 27);
+            trace!("{:width$}type at {}: {}", "", pos, name, width = 0);
         }
     }
 
@@ -47,7 +47,7 @@ impl Constraints {
     /// Only used during unification stage.
     pub fn push(&mut self, msg: &str, left: &Expected, right: &Expected) {
         let constraint = Constraint::new(msg, left, right);
-        trace!("{:width$}[gen {}] {}", "", msg, constraint, width = 2);
+        trace!("{:width$}[gen {}] {}", "", msg, constraint, width = 17);
         self.constraints.push_front(constraint)
     }
 
@@ -69,7 +69,7 @@ impl Constraints {
         if constraint.is_flag {
             // Can only reinsert constraint once
             let msg = format!(
-                "Cannot infer type. Expected {}, was {}",
+                "Cannot infer type. Expected a {}, was {}",
                 &constraint.left.expect, &constraint.right.expect
             );
             return Err(vec![TypeErr::new(constraint.left.pos, &msg)]);
--- a/src/check/constrain/constraint/iterator.rs
+++ b/src/check/constrain/constraint/iterator.rs
@@ -69,7 +69,7 @@ impl Constraints {
         if constraint.is_flag {
             // Can only reinsert constraint once
             let msg = format!(
-                "Cannot infer type. Expected a {}, was {}",
+                "Cannot infer type. Expected {}, was {}",
                 &constraint.left.expect, &constraint.right.expect
             );
             return Err(vec![TypeErr::new(constraint.left.pos, &msg)]);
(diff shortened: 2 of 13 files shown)


tests/system/valid/class.rs
```diff
@@ -1,3 +1,5 @@
+use log::Level;
```
JSAbrahams, line 1: Remove this

JSAbrahams:
Perhaps we are missing a substitute rule for access somewhere, though that does seem unlikely.
Also, it would be nice if we can get rid of all duplicate constraints before going into the unification stage, also for a more succinct trace.
I've tried to check before adding but didn't work yet for some reason.

src/check/constrain/generate/call.rs
```diff
@@ -206,88 +206,72 @@ fn call_parameters(
...
+        _ => return Err(vec![TypeErr::new(property.pos, "Expected property call")]),
+    };
+
+    let entire_call_as_ast: AST = instance.iter().rfold(property.clone(), |acc, ast| {
+        let (instance, property) = (Box::from(ast.clone()), Box::from(acc));
+        AST::new(ast.pos, Node::PropertyCall { instance, property })
+    });
+    let entire_call_as_ast = Expected::try_from((&entire_call_as_ast, &env.var_mappings))?;
+
+    let ast_without_access = if instance.len() == 1 {
```
JSAbrahams, line 256: Bit sad that we don't use a more elegant float but instead an ugly if statement.
But not sure there is another way to do this.

src/check/constrain/constraint/iterator.rs
```diff
@@ -69,7 +69,7 @@ impl Constraints {
         if constraint.is_flag {
             // Can only reinsert constraint once
             let msg = format!(
-                "Cannot infer type. Expected a {}, was {}",
+                "Cannot infer type. Expected {}, was {}",
```
JSAbrahams, line 72: Change these messages back to use `and_or_a`

src/check/constrain/generate/call.rs
```diff
@@ -206,88 +206,72 @@ fn call_parameters(
...
-                            args: args_with_self,
-                        },
-                    )),
-                },
-            );
+            let (constr, _) = gen_vec(args, env, ctx, constr)?;
+            let args = vec![last_inst.clone()]
+                .iter()
+                .chain(args)
+                .map(|ast| Expected::try_from((ast, &env.var_mappings)))
```
JSAbrahams, line 240: This closure could perhaps be on a separate line above

src/check/constrain/generate/call.rs
```diff
@@ -206,88 +206,72 @@ fn call_parameters(
...
+    let entire_call_as_ast: AST = instance.iter().rfold(property.clone(), |acc, ast| {
+        let (instance, property) = (Box::from(ast.clone()), Box::from(acc));
+        AST::new(ast.pos, Node::PropertyCall { instance, property })
+    });
+    let entire_call_as_ast = Expected::try_from((&entire_call_as_ast, &env.var_mappings))?;
+
+    let ast_without_access = if instance.len() == 1 {
+        last_inst.clone()
+    } else {
+        let last = instance.remove(instance.len() - 1);
```
JSAbrahams, line 259: Hm is there a situation where this could panic?

src/check/constrain/unify/expression/substitute.rs
```diff
@@ -22,25 +22,23 @@ pub fn substitute(
...
         constraint_pos += 1;
         macro_rules! replace {
             ($left:expr, $new:expr) => {{
-                let pos =
-                    format!("({}={}) ", old_constr.left.pos, old_constr.right.pos);
                 let side = if $left { "l" } else { "r" };
                 trace!(
                     "{:width$} [subbed {}\\{} {}]  {}  =>  {}",
-                    pos,
+                    "",
```
JSAbrahams, line 35: Surely we still want to print pos somehow?

src/check/constrain/unify/link.rs
```diff
@@ -18,13 +18,12 @@ pub fn unify_link(constraints: &mut Constraints, ctx: &Context, total: usize) ->
...
         let (left, right) = (&constraint.left, &constraint.right);
 
-        let pos = format!("{}={} ", left.pos, right.pos);
         let count = if constraints.len() <= total { total - constraints.len() } else { 0 };
         let unify = format!("{}\\{}", count, total);
         let msg =
             if constraint.msg.is_empty() { String::new() } else { format!(" {}", constraint.msg) };
 
-        trace!("{:width$}[{}{}]  {}", pos, unify, msg, constraint, width = 27);
+        trace!("{:width$}[{}{}]  {}", "", unify, msg, constraint, width = 0);
```
JSAbrahams, line 26: Surely we still want to print pos somehow?

src/check/constrain/constraint/expected.rs
```diff
@@ -128,7 +127,7 @@ impl Display for Expect {
                 Access { entity, name } => format!("{}.{}", entity.expect, name.expect),
                 Function { name, args } => format!("{}({})", name, comma_delm(args)),
                 Field { name } => name.clone(),
-                Type { name: ty } => format!("{}'{}'", an_or_a(ty), ty),
+                Type { name: ty } => format!("{}", ty),
```
JSAbrahams, line 130: This can be a separate PR

src/check/constrain/constraint/iterator.rs
```diff
@@ -34,7 +34,7 @@ impl Constraints {
     pub fn push_ty(&mut self, pos: Position, name: &Name) {
         let name = self.finished.get(&pos).map_or(name.clone(), |s_name| s_name.union(name));
         if self.finished.insert(pos, name.clone()).is_none() {
-            trace!("{:width$}type at {}: {}", "", pos, name, width = 0);
+            trace!("{:width$} type: {}", pos, name, width = 27);
```
JSAbrahams, line 37: This can be a separate PR

src/check/constrain/constraint/iterator.rs
```diff
@@ -47,7 +47,7 @@ impl Constraints {
     /// Only used during unification stage.
     pub fn push(&mut self, msg: &str, left: &Expected, right: &Expected) {
         let constraint = Constraint::new(msg, left, right);
-        trace!("{:width$}[gen {}] {}", "", msg, constraint, width = 17);
+        trace!("{:width$}[gen {}] {}", "", msg, constraint, width = 2);
```
JSAbrahams, line 50: This can be a separate PR

src/check/constrain/constraint/iterator.rs
JSAbrahams: This can be a separate PR

src/check/constrain/unify/expression/substitute.rs
```diff
@@ -22,25 +22,23 @@ pub fn substitute(
     let mut substituted = Constraints::new();
     let mut constraint_pos = offset;
 
-    trace!("{:width$} [subbing {}\\{}]  {}  <=  {}", "", offset, total, old, new, width = 30);
+    trace!("{:width$} [subbing {}\\{}]  {}  <=  {}", "", offset, total, old, new, width = 10);
```
JSAbrahams, line 25: This can be a separate PR

src/check/constrain/unify/expression/substitute.rs
JSAbrahams: This can be a separate PR

src/check/constrain/unify/link.rs
JSAbrahams: This can be a separate PR

src/check/constrain/unify/link.rs
```diff
@@ -74,7 +73,7 @@ pub fn unify_link(constraints: &mut Constraints, ctx: &Context, total: usize) ->
 pub fn reinsert(constr: &mut Constraints, constraint: &Constraint, total: usize) -> Unified {
     let pos = format!("({}={}) ", constraint.left.pos.start, constraint.right.pos.start);
     let count = format!("[reinserting {}\\{}] ", total - constr.len(), total);
-    trace!("{:width$}{}{}", pos, count, constraint, width = 17);
+    trace!("{:width$}{}{}", pos, count, constraint, width = 15);
```
JSAbrahams, line 76: This can be a separate PR

src/check/result.rs
```diff
@@ -103,6 +103,9 @@ impl WithSource for TypeErr {
 impl Display for TypeErr {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         let path = self.path.clone().map_or(String::from("<unknown>"), |p| p.display().to_string());
+        // remove trailing slash
+        let path = path.strip_suffix(MAIN_SEPARATOR).unwrap_or(&path);
```
JSAbrahams, line 107: This can be a separate PR

src/generate/result.rs
```diff
@@ -55,10 +55,14 @@ impl WithSource for UnimplementedErr {
 
 impl Display for UnimplementedErr {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        let path =
+            self.path.clone().map_or(String::from("<unknown>"), |path| path.display().to_string());
+        let path = path.strip_suffix(MAIN_SEPARATOR).unwrap_or(&path);
```
JSAbrahams, line 60: This can be a separate PR

src/parse/result.rs
```diff
@@ -1,7 +1,7 @@
 use std::cmp::min;
 use std::fmt;
 use std::fmt::{Display, Formatter};
-use std::path::PathBuf;
+use std::path::{MAIN_SEPARATOR, PathBuf};
```
JSAbrahams, line 4: This can be a separate PR

src/parse/result.rs
```diff
@@ -173,6 +173,9 @@ impl Display for ParseErr {
                 )
             });
 
+        let path =
+            self.path.clone().map_or(String::from("<unknown>"), |path| path.display().to_string());
+        let path = path.strip_suffix(MAIN_SEPARATOR).unwrap_or(&path);
```
JSAbrahams, line 178: This can be a separate PR

PR #341, opened 2022-05-31
Context restructure

### LINKED ISSUE #321: Function return is not super of body when body nullable
### How to Reproduce

```
    def some_higher_order(self, fun: Int -> Int) -> Int? => fun(10)
    def fancy(self) -> Int => self.some_higher_order(\x: Int => x * 2)
```

### Expected behavior

The above should fail because body is nullable, whereas the function signature does say that the return type is integer.

JSAbrahams:
### Relevant issues

- Closes #321 
- Made note of #343 

### Summary

- Generalize how class lookup work in `Context`

No major bugs have really been fixed.
However, we have restructured the Context so we don't have to deal with a lot of edge cases and instead use a more generalized system.

### Added Tests

- Print `Tuple` and `Callable`.
- Some checks that `HasParent` has been properly implemented for some primitives.


tests/resource/invalid/type/definition/function_ret_not_super.mamba
```diff
@@ -0,0 +1,3 @@
+def f(x: Int) -> Int => 10
```
JSAbrahams, line 1: Remove

src/check/name/namevariant.rs
```diff
@@ -16,12 +17,38 @@ pub enum NameVariant {
...
+            (NameVariant::Tuple(l_name), NameVariant::Tuple(r_name)) => l_name.partial_cmp(r_name),
+            (NameVariant::Fun(l_args, l_ret), NameVariant::Fun(r_args, r_ret)) => {
+                let cmp = l_args.partial_cmp(r_args);
+                if let Some(Ordering::Equal) = cmp {
+                    l_ret.partial_cmp(r_ret)
+                } else {
+                    cmp
+                }
+            }
+            _ => todo!(),
```
JSAbrahams, line 35: Finish this function

src/check/context/mod.rs
```diff
@@ -183,94 +169,69 @@ impl LookupField<&str, Field> for Context {
...
-        }
-    }
-
-    pub fn name(&self) -> TrueName {
-        let variant = match &self.variant {
-            ClassVariant::Direct(class) => return class.name.clone(),
-            ClassVariant::Tuple(classes) => {
-                NameVariant::Tuple(classes.iter().map(|c| c.name()).collect())
+impl ClassVariant {
+    pub fn field(&self, name: &str, ctx: &Context, pos: Position) -> TypeResult<FieldUnion> {
```
JSAbrahams, line 185: This should be a trait

src/check/context/mod.rs
```diff
@@ -183,94 +169,69 @@ impl LookupField<&str, Field> for Context {
...
-
-                    let variant = NameVariant::Tuple(classes.iter().map(|c| c.name()).collect());
-                    let self_arg = Name::from(&TrueName::from(&variant));
-                    let ret_ty = Name::from(clss::STRING_PRIMITIVE);
-                    Function::simple_fun(name, &self_arg, &ret_ty, pos)
-                } else {
-                    let msg = format!("Function '{}' undefined on '{}'", name, self);
-                    Err(vec![TypeErr::new(pos, &msg)])
-                }
+    pub fn fun(&self, name: &StringName, ctx: &Context, pos: Position) -> TypeResult<FunUnion> {
```
JSAbrahams, line 201: This should be a trait

src/check/context/mod.rs
```diff
@@ -360,48 +355,172 @@ impl HasParent<&Name> for ClassUnion {
 
 impl ClassUnion {
```
JSAbrahams, line 356: Move ClassUnion to its own file

src/check/context/mod.rs
```diff
@@ -360,48 +355,172 @@ impl HasParent<&Name> for ClassUnion {
...
             return Err(vec![TypeErr::new(pos, &msg)]);
         }
 
-        Ok(FieldUnion { union })
+        Ok(FieldUnion::from(&union))
     }
 }
 
+#[derive(Debug, Eq)]
 pub struct FunUnion {
```
JSAbrahams, line 431: Move FunUnion to its own file

src/check/context/mod.rs
```diff
@@ -360,48 +355,172 @@ impl HasParent<&Name> for ClassUnion {
...
+            Ok(self.union.iter().next().unwrap().clone())
+        } else {
+            let msg = format!("Expected single function but was {}", &self);
+            Err(vec![TypeErr::new(pos, &msg)])
+        }
+    }
+}
+
+#[derive(Debug, Eq)]
 pub struct FieldUnion {
```
JSAbrahams, line 478: Move to its own file

src/check/context/mod.rs
```diff
@@ -109,20 +95,20 @@ impl LookupClass<&StringName, Class> for Context {
     }
 }
 
-impl LookupClass<&TrueName, ClassTuple> for Context {
-    fn class(&self, class: &TrueName, pos: Position) -> TypeResult<ClassTuple> {
-        let variant = match &class.variant {
-            NameVariant::Single(direct) => ClassVariant::Direct(self.class(direct, pos)?),
+impl LookupClass<&TrueName, ClassVariant> for Context {
```
JSAbrahams, line 98: Think all lookups can be grouped together in a file

src/check/context/mod.rs
```diff
@@ -183,94 +169,69 @@ impl LookupField<&str, Field> for Context {
     }
 }
 
-#[allow(clippy::large_enum_variant)]
-#[derive(Debug, PartialEq, Eq, Hash)]
+#[derive(Debug, Eq)]
 pub enum ClassVariant {
```
JSAbrahams, line 173: ClassVariant and all associated logic could at the very least be in its own file

JSAbrahams (commented):
Split up logic into more files for readability

tests/resource/valid/definition/function_ret_super.mamba
```diff
@@ -0,0 +1,2 @@
+def some_higher_order(fun: Int -> Int) -> Int => fun(10)
+def fancy() -> Int? => some_higher_order(\x: Int => x * 2)
```
JSAbrahams, line 2: Question is, why did this not fail before? We just copied this from another test, but now `fancy` does not have argument `self`, and is not within a class.
Should also add a test with these exact functions but within a class to see if that type checks.

JSAbrahams (commented):
A test which is very similar to the one before for some reason fails while the other previously didn't. So this may be indicative that the checker is not checking something.

src/check/context/clss/concrete/variant.rs
```diff
@@ -0,0 +1,167 @@
...
+    fn fun(&self, name: &StringName, ctx: &Context, pos: Position) -> TypeResult<FunUnion> {
+        match &self {
+            ClassVariant::Direct(class_set) => {
+                let funs: HashSet<Function> = class_set
+                    .iter()
+                    .map(|class| class.fun(name, ctx, pos))
+                    .collect::<Result<_, _>>()?;
+                Ok(FunUnion::from(&funs))
+            }
+            other => {
```
JSAbrahams, line 131: We can do this more elegantly by treating `ClassVariant::Tuple` as a `tuple` class, for which Python does have some methods:
`index`, `count`, `__str__`, etc.

So we just call that class and treat this as though one called `ClassVariant::Direct(class_set = {Tuple})`

JSAbrahams: Same for `ClassVariant::Fun`, python already defines some methods for this class, such as `__str__`

JSAbrahams (commented):
How to fix calls to tuple and callable, at the very least make them stringy

src/check/name/namevariant.rs
JSAbrahams: Order: Single, Tuple, Fun

JSAbrahams:
The drop in coverage is somewhat significant.
A lot of logic has been moved around or added.
A lot of it is not really that necessary to test though, since a lot of logic is about the order of names.
This is mostly so that the ordering is consistent for cases where we need to e.g. order sets in a deterministic manner.

PR #352, opened 2022-06-10
Streamline Name logic

### LINKED ISSUE #335: Test `Context` `GenericFunction` and the like
Add some extra unit tests for these just as a final check before moving on.

JSAbrahams:
### Relevant issues

- Closes #335 

Might fix some other underlying issues due to streamlining of logic.

### Summary

- Make Tuple and Callable interchangeable with a single name in certain situations.
- Substitute generics when looking up parent.

### Added Tests

- `Tuple` has parent `Collection[MyGeneric]`.
- Same test but if tuple happens to be a name variant.
- `Callable` has parent `Callable`.
- Test that Callable does not have Callable as parent with wrong arg or ret type


src/check/context/clss/concrete/mod.rs
```diff
@@ -1,2 +1,56 @@
...
+        let pos = Position::default();
+        let tuple = ctx.class(&Name::from(clss::TUPLE), pos)?;
+
+        let collection = StringName::new(clss::COLLECTION, &[]);
+        assert!(tuple.has_parent(&TrueName::from(&collection), &ctx, pos)?);
+        Ok(())
+    }
+
+    #[test]
+    fn namevariant_tuple_has_collection_parent() -> TypeResult<()> {
```
JSAbrahams, line 28: Get these tests working ASAP

src/check/context/clss/mod.rs
```diff
@@ -138,23 +126,22 @@ impl LookupClass<&StringName, Class> for Context {
     ///
     /// Substitutes all generics in the class when found.
     fn class(&self, class: &StringName, pos: Position) -> TypeResult<Class> {
-        if let Some(generic_class) = self.classes.iter().find(|c| {
-            StringName::try_from_pos(&c.name, pos).map(|name| name.name) == Ok(class.name.clone())
-        }) {
+        if let Some(generic_class) = self.classes.iter().find(|c| c.name.name == class.name.clone())
```
JSAbrahams, line 129: Can't we just use a reference here?

src/check/context/clss/concrete/mod.rs
```diff
@@ -1,2 +1,56 @@
...
+        ])));
+        let tuple = ctx.class(&name, pos).expect("Tuple");
+
+        let collection = StringName::new(clss::COLLECTION, &[Name::from("GENERIC"), Name::from("GENERIC")]);
+        assert!(tuple.has_parent(&TrueName::from(&collection), &ctx, pos)?);
+        Ok(())
+    }
+
+    #[test]
+    fn namevariant_callabel_has_callable_parent() -> TypeResult<()> {
```
JSAbrahams, line 43: Add generics to internal `Callable` of type checker

src/check/context/clss/concrete/variant.rs
```diff
@@ -32,10 +32,10 @@ impl From<&ClassVariant> for Name {
     fn from(class_variant: &ClassVariant) -> Self {
         let name_variant = match &class_variant {
             ClassVariant::Direct(class) if class.len() == 1 => {
-                class.iter().next().unwrap().name.variant.clone()
+                NameVariant::Single(class.iter().next().unwrap().name.clone())
             }
             ClassVariant::Direct(classes) => {
```
JSAbrahams, line 37: Not strictly speaking part of this PR, but why do we even differentiate between these two cases?

src/check/context/clss/concrete/variant.rs
```diff
@@ -32,10 +32,10 @@ impl From<&ClassVariant> for Name {
...
-                let names = classes.iter().map(|class| &class.name.variant).map(Name::from);
+                let names = classes.iter().map(|class| &class.name).map(Name::from);
                 return Name::from(&HashSet::from_iter(names));
             }
             ClassVariant::Tuple(class_unions) => {
```
JSAbrahams, line 41: Add tests for these branches

PR #353, opened 2022-06-17
ABC

### LINKED ISSUE #331: Interfaces should specify that they are a `ABC` (Abstract Base Class)
### Current Issue

Types are just treated as classes.

Python has a package called `abc` with, among other things:

- `ABC`, Abstract Base Class.
- `abstractmethod`, which speaks for itself.

Generated python code should specify for interfaces that these are abstract by making use of the above.

JSAbrahams:
### Relevant issues

- Partially fixes #331

### Summary

- Group from imports together
- Specify @abstractmethod
- Specify ABC class if abstract.
  However, also added to classes which already
  have a parent which is abstract.
  This redundancy can be fixed in the next PR.

However, generate stage does not use a Context to check whether an abstract class already had a parent which is abstract. 
This results in redundant `ABC` parents.
This should be also fixed (in a separate PR) before closing #331.

### Added Tests
- Update some test to check that `ABC` and `@abstractmethod` imported and used.
- Concrete method in abstract class (interface)

### DIFF
--- a/src/generate/mod.rs
+++ b/src/generate/mod.rs
@@ -80,11 +80,13 @@ pub fn gen_arguments(ast_ty: &ASTTy, gen_args: &GenArguments) -> GenResult {
     match convert_node(ast_ty, import, &state)? {
         Core::Block { statements: mut old_stmts } => {
             let mut statements = import.imports.clone();
+            statements.append(&mut import.from_imports.clone().into_values().collect());
             statements.append(&mut old_stmts);
             Ok(Core::Block { statements })
         }
         other if !import.imports.is_empty() => {
             let mut statements = import.imports.clone();
+            statements.append(&mut import.from_imports.clone().into_values().collect());
             statements.push(other);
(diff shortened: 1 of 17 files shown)


src/generate/mod.rs
```diff
@@ -80,11 +80,13 @@ pub fn gen_arguments(ast_ty: &ASTTy, gen_args: &GenArguments) -> GenResult {
...
         Core::Block { statements: mut old_stmts } => {
             let mut statements = import.imports.clone();
+            statements.append(&mut import.from_imports.clone().into_values().collect());
             statements.append(&mut old_stmts);
             Ok(Core::Block { statements })
         }
         other if !import.imports.is_empty() => {
             let mut statements = import.imports.clone();
+            statements.append(&mut import.from_imports.clone().into_values().collect());
             statements.push(other);
```
JSAbrahams, line 90: Create a single `imports()` method which then does this appending internally

PR #355, opened 2022-06-24
Add bytes primitive

### LINKED ISSUE #332: Add python `bytes` primitive to language

JSAbrahams:
### Relevant issues

- Closes #332


JSAbrahams:
Note that actual python does not support function overloading.
So, if we want that we should make clear that the internal resources of the type checker are a sort of pseudo-Python.
We can then use this overloading to basically condition the output type of a function on the number of arguments and types of arguments to the same function.

JSAbrahams:
We now also need a system where we can say "here is a set of constraints, at least one of them should type check" for situations where we have multiple alternatives and we cannot locally determine which one would check.

Issue #296, opened 2022-05-23
Type checker does not take largest demoninator


JSAbrahams:
### Description of Bug

Type checker does not take largest demoninator.

### How to Reproduce

```
def a := 10

a += 2
a -= 3
a *= 6

a /= 7
a ^= 2

a <<= 10
a >>= 5
```

### Expected behavior

Results in type checker tripping on `a`, should be `Float` but checker assumes it is `Int`.

JSAbrahams:
Lets move to next milestone since this might require a non-trivial restructuring.

PR #360, opened 2022-11-17
Automatically convert last expression to return

### LINKED ISSUE #164: Functions with return type should return last expression
### Current Issue
Unless we explicitly end a function with `return`, it does not return anything.

In Mamba, a function that ends with an expression returns that type as long if the signature states as such.

See #165

JSAbrahams:
### Relevant issues

Resolves #164 

### Summary

Function with return last item in function body.
We do this by making use of the state in the generate stage.
When a function has a return in the signature, a flag is set in the state.
This flag is then used in the generate of the body, but it is not propagated further to prevent returns being inserted deep within the AST where they shouldn't be.
If set to true, then we in most situations wrap the final item of the function body in a core return node.
Exceptions are explained below.

If function body does not end with an expression, then the check stage already gives an error (we tested this before, right? Maybe check).
We have some exceptions:
- If last item is a return, which the check stage does allow (meaning we must deal with in the generate stage), we do not generate an additional return.
- If the last item is an if-else, then we do not insert a return here (which would result in incorrect syntax), and instead propagate forward the state to the branches if the if-else. We take care not to propagate into the condition of the if-else.
- In all other cases, an expression is wrapped by a return core node

Also exposed a weird error in check stage. 
In `function/definition.mamba`, line 7:
Change
`    if True then return 10 else return None`
To
`    if True then 10 else None`
Gives type error:
`Unifying two types: Expected an Int, was a None`
Not something to fix in this PR, it is erroneous behaviour as the mamba script is well-typed according to current spec.

### Added Tests

- Check that function with statement and expression return last statement
- Modified current tests to conform to new behaviour

### DIFF
--- a/src/generate/convert/control_flow.rs
+++ b/src/generate/convert/control_flow.rs
@@ -9,7 +9,7 @@ pub fn convert_cntrl_flow(ast: &ASTTy, imp: &mut Imports, state: &State, ctx: &C
     Ok(match &ast.node {
         NodeTy::IfElse { cond, then, el } => match el {
             Some(el) => Core::IfElse {
-                cond: Box::from(convert_node(cond, imp, state, ctx)?),
+                cond: Box::from(convert_node(cond, imp, &state.last_ret(false), ctx)?),
                 then: Box::from(convert_node(then, imp, state, ctx)?),
                 el: Box::from(convert_node(el, imp, state, ctx)?),
             },
--- a/src/generate/convert/mod.rs
+++ b/src/generate/convert/mod.rs
@@ -278,12 +278,12 @@ pub fn convert_node(ast: &ASTTy, imp: &mut Imports, state: &State, ctx: &Context
         match core {
             Core::Block { ref statements } => if let Some(last) = statements.last() {
                 match last {
-                    Core::Return { .. } => core.clone(),
+                    last if skip_return(last) => core.clone(),
                     _ => {
                         let mut new_stmts = statements.clone();
-                        let last = if let Some(last) = new_stmts.pop() {
+                        let last = if let Some(ref last) = new_stmts.pop() {
                             match last {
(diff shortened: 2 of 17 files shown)


JSAbrahams:
The "weird error" in the check stage, at first glance, would appear to be related to #296

src/generate/convert/mod.rs
```diff
@@ -271,6 +274,32 @@ pub fn convert_node(ast: &ASTTy, imp: &mut Imports, state: &State, ctx: &Context
...
+                        };
+
+                        new_stmts.push(last);
+                        Core::Block { statements: new_stmts }
+                    }
+                }
+            } else {
+                core.clone()
+            }
+            Core::Return { .. } | Core::IfElse { .. } | Core::Raise { .. } => core,
```
JSAbrahams, line 300: Are we absolutely sure these are the only constructs?
Best to make this a named function as well for readability.
I'm guessing that stuff like `for loop`s and `while loop`s should not have a return (Even if they won't pass the check stage if a function ends with those).

src/generate/convert/control_flow.rs
```diff
@@ -9,7 +9,7 @@ pub fn convert_cntrl_flow(ast: &ASTTy, imp: &mut Imports, state: &State, ctx: &C
     Ok(match &ast.node {
         NodeTy::IfElse { cond, then, el } => match el {
             Some(el) => Core::IfElse {
-                cond: Box::from(convert_node(cond, imp, state, ctx)?),
+                cond: Box::from(convert_node(cond, imp, &state.last_ret(false), ctx)?),
```
JSAbrahams, line 12: Feels a bit finicky to add logic here for the if-else construct (though it is the only construct where the `last_ret` is propagated, see `convert_node` fun).

JSAbrahams: I guess the `if-else` is the only branching language construct, so it makes sense that this is the only place that this happens. Still a bit weird to have it located here but I can't think of another solution.

PR #364, opened 2022-11-23
Ternary expressions

### LINKED ISSUE #164: Functions with return type should return last expression

JSAbrahams:
### Relevant issues

Closes #164
Should've been a separate issue but couldn't be bothered to create one.

Introduced bug highlighted in #366 .
Discovered bug highlighted in #367

### Summary

Treat if expressions as expressions where relevant by converting them to python ternary expressions.

- Remove `ExpressionAny`, which introduced a lot of unnecessary logic and complexity.
  Streamlining of logic made it easier to implement this new feature.
  Instead, just have a dedicated `Any` type and match this if needed.
  (Perhaps this introduced the bug in #366?).

### Added Tests

- *Happy* assign propagated into if
- *Happy* ... and match
- *Happy* If converted to ternary when used as function argument

### DIFF
--- a/src/check/constrain/generate/mod.rs
+++ b/src/check/constrain/generate/mod.rs
@@ -82,6 +82,7 @@ pub fn generate(
         Mod { .. } => gen_op(ast, env, ctx, constr),
         AddU { .. } | SubU { .. } => gen_op(ast, env, ctx, constr),
         Sqrt { .. } => gen_op(ast, env, ctx, constr),
+        Undefined => gen_op(ast, env, ctx, constr),
 
         BOneCmpl { .. } => gen_op(ast, env, ctx, constr),
         BAnd { .. } | BOr { .. } | BXOr { .. } => gen_op(ast, env, ctx, constr),
@@ -103,7 +104,6 @@ pub fn generate(
         | ExpressionType { .. }
         | DocStr { .. }
         | Underscore
-        | Undefined
         | Comment { .. } => Ok((constr.clone(), env.clone())),
     }
 }
--- a/src/check/constrain/generate/operation.rs
+++ b/src/check/constrain/generate/operation.rs
@@ -2,8 +2,8 @@ use std::convert::TryFrom;
 
 use crate::check::constrain::constraint::builder::ConstrBuilder;
 use crate::check::constrain::constraint::Constraint;
-use crate::check::constrain::constraint::expected::{Expect, Expected};
 use crate::check::constrain::constraint::expected::Expect::*;
+use crate::check::constrain::constraint::expected::Expected;
 use crate::check::constrain::generate::{Constrained, gen_vec, generate};
 use crate::check::constrain::generate::collection::{constr_col, gen_collection_lookup};
 use crate::check::constrain::generate::env::Environment;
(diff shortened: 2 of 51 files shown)


JSAbrahams:
Before merging, it'd be nice actually to check other situations where an if expression should be converted to a Python ternary expression.
Currently, we only do this within a function argument.
What about using an (inline-)if expressions for:
- Variable declarations
- Nested further within other if's
- Used as a condition for an if, match
- Used in any other location as an expression, e.g. a for loop condition

JSAbrahams:
Before merging, it'd be nice actually to check other situations where an if expression should be converted to a Python ternary expression.
Currently, we only do this within a function argument.
What about using an (inline-)if expressions for:
- Variable declarations
- Nested further within other if's
- Used as a condition for an if, match
- Used in any other location as an expression, e.g. a for loop condition

src/check/constrain/generate/mod.rs
```diff
@@ -82,6 +82,7 @@ pub fn generate(
         Mod { .. } => gen_op(ast, env, ctx, constr),
         AddU { .. } | SubU { .. } => gen_op(ast, env, ctx, constr),
         Sqrt { .. } => gen_op(ast, env, ctx, constr),
+        Undefined => gen_op(ast, env, ctx, constr),
```
JSAbrahams, line 85: Note for #366 to look here.

src/check/constrain/generate/operation.rs
```diff
@@ -116,24 +123,24 @@ pub fn gen_op(
 
         Node::BOneCmpl { expr } => {
             let left = Expected::try_from((expr, &env.var_mappings))?;
-            constr.add("binary compliment", &left, &Expected::new(expr.pos, &ExpressionAny));
+            constr.add("binary compliment", &left, &Expected::new(expr.pos, &Type { name: Name::any() }));
```
JSAbrahams, line 126: What if we have a function for type so we can have `Expect::Any`. 
Implement `Any` interface for `Expect`?

src/check/constrain/generate/resources.rs
```diff
@@ -32,7 +32,7 @@ pub fn gen_resources(
             constr.new_set(true);
             let resource_exp = Expected::try_from((resource, &env.var_mappings))?;
             constr.add("with as", &resource_exp, &Expected::try_from((alias, &env.var_mappings))?);
-            constr.add("with as", &resource_exp, &Expected::new(resource.pos, &ExpressionAny));
+            constr.add("with as", &Expected::new(resource.pos, &Type { name: Name::any() }), &resource_exp);
```
JSAbrahams, line 35: See comment about `Expect::any()`

src/generate/convert/control_flow.rs
```diff
@@ -7,26 +7,49 @@ use crate::generate::result::{GenResult, UnimplementedErr};
...
-                then: Box::from(convert_node(then, imp, state, ctx)?),
-                el: Box::from(convert_node(el, imp, state, ctx)?),
-            },
-            None => Core::If {
-                cond: Box::from(convert_node(cond, imp, state, ctx)?),
-                then: Box::from(convert_node(then, imp, state, ctx)?),
-            },
-        },
+        NodeTy::IfElse { cond, then, el } => {
+            let cond = Box::from(convert_node(cond, imp, &state.is_last_must_be_ret(false).must_assign_to(None), ctx)?);
```
JSAbrahams, line 11: A bit sloppy to have to reset state like this constantly, but I can't think of another way to do this.

JSAbrahams:
Looking at codecov, few notes:
- Coverage drops because we have a lot if where the else block is a panic.
  Can we use `let-else` instead: https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html?
  Though this doesn't actually fix the panic issue.
- Unification of collection and type is never tested.
  Do we even need to be able to recognize this union?
  Maybe we do, but if we don't this is just unnecessary logic.

PR #369, opened 2022-12-17
Remove constraint variant either

### LINKED ISSUE #366: Check doesn't notice if fun returns `None` for non-nullable ret type
### Description of Bug

Type check does not catch if function returns `None` when it should return a non-nullable type.
At the very least, it doesn't catch this if the `None` is nested in a ternary expression.

### How to Reproduce

Run a check on:
`def f() -> Int => if True then 10 else None`

### Expected behavior

Should not check.

JSAbrahams:
### Relevant issues

Resolves #366 
Should make an issue to discuss whether we want to allow different types for arms of match and if expressions: #370 

### Summary
This was redundant and added unnecessary logic I believe. 
One problem is that now that the arms of an if expression must always match. 
One cool feature would be that the language then instead takes a union of these two types and continues on.

### Added Tests

- *Happy* Add test to check assign isn't overridden, more relevant to #364

### DIFF
--- a/src/check/constrain/constraint/mod.rs
+++ b/src/check/constrain/constraint/mod.rs
@@ -24,15 +24,13 @@ pub struct Constraint {
 pub enum ConstrVariant {
     Left,
     Right,
-    Either,
 }
 
 impl Display for Constraint {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), Error> {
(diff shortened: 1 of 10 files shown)


JSAbrahams:
Well, the reduction is coverage is unfortunate. But not much we can do about it in this PR I think.

PR #372, opened 2022-12-17
Version 0.3.4

JSAbrahams:
### Relevant issues

Closes milestone 0.3.4

### Summary

- Add if expressions, which are converted to Python ternary expressions
- Allow match expressions (i.e. match to be treated as an expression)
- For both if and match expressions:
  - Convert expressions in if and match arms to return in function if necessary
  - Propagate assign to variable inwards into arms if necessary
- Check stage checks if body of function was empty
- Interfaces specify that they are `ABC` in Python (Abstract Base Class)
- `None` is treated as a special class, not as a special keyword


JSAbrahams:
Realised that adding `return` and assigns should be done recursively using a mapping function instead of doing it top-level.
Will add in a PR soon.

JSAbrahams:
Bit unfortunate to have a commit without associate PR in main, but lets leave it as-is.

PR #377, opened 2022-12-21
Fix build and coverage badge

JSAbrahams:
### Summary

- Fix build badge, now properly displayed
- Specify that coverage must be run for each new release.


JSAbrahams:
I at least hope that codecov does update the coverage when we have a new release

Issue #375, opened 2022-12-21
Check does not notice erroneous types when assigning using nested ifs


JSAbrahams:
### Description of Bug

Check does not notice erroneous types when assigning using nested ifs.

### How to Reproduce

```
def x := 20
def a: int := if x > 10 then
    if x > 30 then
        "a string"
    else
        x - 1
else
    x
```

### Expected behavior

Should fail because `a` has type `Int` even though there is a possibility it could be a string.

PR #380, opened 2022-12-22
Generate identifiers from conditions of match arms

JSAbrahams:
### Summary

Match arms treated same way as if arms:
- For each new arm create new constraint set
- Use similar logic to for loop variable initialisation.

In future should have way to have some sort of union between constraints, much like env.
This allows us to type check all possible execution paths.

When we implementing pattern matching this system will need to be extended.

### Added Tests

*Happy*
- Variable only declared in condition of match arm usable in arm body.

*Sad*
- Shadow system does not delete outer variables of match.
  Started writing this while creating system of environment unions between match body arms.
  However, I don't think that we actually want this behaviour.

### DIFF
--- a/tests/resource/invalid/type/control_flow/access_match_arms_variable.mamba
+++ b/tests/resource/invalid/type/control_flow/access_match_arms_variable.mamba
@@ -1,0 +1,5 @@
+match 10
+    n =>
+        print("10")
+
+print(n)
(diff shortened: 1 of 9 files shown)


tests/resource/invalid/type/control_flow/access_match_arms_variable.mamba
```diff
@@ -0,0 +1,5 @@
+match 10
+    n =>
+        print("10")
```
JSAbrahams, line 3: What about a test where we use an undefined variable in match arm body
Also one where we shadow a variable, but use it in the same manner as we would've the outer variable (e.g. assuming it is an `Int` while it is a `String`)

JSAbrahams: No system for the second, though this is somewhat related to #375

PR #382, opened 2022-12-22
Handle exceptions

### LINKED ISSUE #374: Exceptions in handle arms not taken into account
### Description of Bug

Type checker cannot check whether function returns exception.

### How to Reproduce

```
def g() -> Int raise [Exception] =>
    raise Exception("A")

def f(x: Int) -> Int=>
    g() handle
        err: Exception => x + 10
```

Gives the following error:
```
Error: Exceptions not covered: Exception
 --> my_file.mamba:5:5
   4 | def f(x: Int) -> Int =>
   5 |     g() handle
           ^^^
   6 |         err: Exception => x + 10
```

### Expected behavior

Should pass check stage.
The handle does have a match arm for the `Exception` error.

JSAbrahams:
### Relevant issues

- Do not merge before #380 
- Resolves #374 
- Created #383 
- Created #385 
- Created #386 
### Summary

Fix handle:
- Checks that handle arms cover all exceptions.
- Also allow that (only) parents are covered of exceptions.
- Current logic allows handle cover more exceptions than may be raised.

### Added Tests

**Happy**
- Handle
- Handle cover only parent of exception
- Exception in function no covered but in function signature

**Sad**
- Using undefined variable in match arm body
- Exception in function not covered

### DIFF
--- a/src/check/constrain/constraint/expected.rs
+++ b/src/check/constrain/constraint/expected.rs
@@ -69,7 +69,6 @@ pub enum Expect {
     Expression { ast: AST },
     Collection { ty: Box<Expected> },
     Tuple { elements: Vec<Expected> },
-    Raises { name: Name },
     Function { name: StringName, args: Vec<Expected> },
     Field { name: String },
     Access { entity: Box<Expected>, name: Box<Expected> },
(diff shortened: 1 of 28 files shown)


JSAbrahams:
Should test that error raised when exception is not handled

PR #387, opened 2022-12-27
Treat constraint builder as mutable reference

### LINKED ISSUE #375: Check does not notice erroneous types when assigning using nested ifs

### LINKED ISSUE #385: ConstrBuilder in check treated both as immutable and mutable argument
# Summary of current issue

It's a bit odd that we in the check stage both:
- Return the constr (`ConstrBuilder`), which is functional, since doing this assumes the original argument is immutable (meaning that we implicitly modify a local clone)
- Pass the constr as a mutable reference (`&mut ConstrBuilder`) and modify this directly.

It is sloppy to use both.
Noticed this when writing the `constraint_cases` function in #382 

# Proposed change

It would be detrimental to performance I think to create a copy each time.
So I would propose that we modify the `Constrained` type to only include the environment.
The environment should be immutable as we don't want any chance of contamination.
In the case of the `ConstrBuilder`, it internally maintains several distinct sets of constraints to prevent contaminiation.

tl;dr:
- Change `Constrained` from:
  `pub type Constrained<T = (ConstrBuilder, Environment)> = Result<T, Vec<TypeErr>>;`
  to:
  `pub type Constrained<T = Environment> = Result<T, Vec<TypeErr>>;`
- Change returns where relevant (lots of places)

JSAbrahams:
### Relevant issues

Resolves #385 
Resolves #375 

### Summary

Simplify logic by treating constraint builder as a mutable argument, without cloning it ever.
Also make sure to not unnecessarily copy over Environment, preventing cross-contamination.

As a happy bonus, this refactor also fixed #375 .


JSAbrahams:
Lets see if we can't uncommon some tests to see if we have unknowingly solved old bugs by simplifying logic.

src/check/constrain/generate/class.rs
```diff
@@ -26,12 +26,12 @@ pub fn gen_class(
...
                 _ => Err(vec![TypeErr::new(body.pos, "Expected code block")])
             },
-        Node::Class { .. } | Node::TypeDef { .. } => Ok((constr.clone(), env.clone())),
+        Node::Class { .. } | Node::TypeDef { .. } => Ok(env.clone()),
 
         Node::TypeAlias { conditions, isa, .. } => constrain_class_body(conditions, isa, env, ctx, constr),
         Node::Condition { cond, el: Some(el) } => {
-            let (mut constr, env) = generate(cond, env, ctx, constr)?;
-            generate(el, &env, ctx, &mut constr)
+            let env = generate(cond, env, ctx, constr)?;
```
JSAbrahams, line 33: `env` should not be preserved here

PR #389, opened 2022-12-27
Fix Readme examples

JSAbrahams:
### Summary

Update README examples

We also (accidentally here) update the bool primitive constructor so that it accepts either a `Bool`, `Int`, or `String`.

### Added Tests

*Happy*
- README examples

### DIFF
--- a/src/check/resource/primitive/bool.py
+++ b/src/check/resource/primitive/bool.py
@@ -1,5 +1,5 @@
 class bool:
-    def __init__(self, arg: bool) -> bool: pass
+    def __init__(self, arg: Union[int, str, bool]) -> bool: pass
 
     def __bool__(self) -> bool: pass
     def __str__(self) -> str: pass
--- a/README.md
+++ b/README.md
@@ -92,17 +92,21 @@ class MyServer(def ip_address: IPv4Address)
     def is_connected: Bool     := False
     def _last_message: String? := None
 
-    def last_sent(fin self) -> String raise ServerError => if self._last_message /= None 
-        then self._last_message
-        else raise ServerError("No last message!")
+    def last_sent(fin self) -> String raise [ServerError] =>
+        if self._last_message != None then
+            self._last_message
+        else
+            raise ServerError("No last message!")
 
     def connect(self) =>
         self.is_connected := True
         print(always_the_same_message)
 
-    def send(self, message: String) raise ServerError => if self.is_connected 
-        then self._last_message := message
-        else raise ServerError("Not connected!")
+    def send(self, message: String) raise [ServerError] =>
+        if self.is_connected then
+            self._last_message := message
+        else
+            raise ServerError("Not connected!")
 
     def disconnect(self) => self.is_connected := False
 ```
@@ -151,9 +155,11 @@ class MyServer(self: DisConnMyServer, def ip_address: IPv4Address): Server
     def is_connected: Bool     := False
     def _last_message: String? := None
 
-    def last_sent(self) -> String raise ServerErr => if self.last_message /= None 
-        then self._last_message
-        else raise ServerError("No last message!")
+    def last_sent(self) -> String raise ServerErr => 
+        if self.last_message != None then 
+            self._last_message
+        else
+            raise ServerError("No last message!")
 
     def connect(self: DisConnMyServer) => self.is_connected := True
 
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ def factorial(x: Int) -> Int => match x
 
 def num := input("Compute factorial: ")
 if num.is_digit() then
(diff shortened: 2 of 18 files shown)


src/check/constrain/generate/control_flow.rs
```diff
@@ -39,7 +39,7 @@ pub fn gen_flow(
                 .raises_caught(&raises_before);
 
             constrain_cases(ast, cases, &outer_env, ctx, constr)?;
-            Ok(env.clone())
+            Ok(outer_env.clone())
```
JSAbrahams, line 42: This logic should be implemented in it's own PR.
This was to make sure that variables defined in handle can be used after handle.

src/check/constrain/generate/operation.rs
```diff
@@ -159,12 +160,16 @@ pub fn gen_op(
             generate(left, env, ctx, constr)?;
             Ok(env.clone())
         }
-        Node::IsA { left, right } | Node::IsNA { left, right } => {
-            let bool = Expected::new(ast.pos, &Type { name: Name::from(clss::BOOL) });
-            constr.add("and", &Expected::try_from((ast, &env.var_mappings))?, &bool);
-            generate(right, env, ctx, constr)?;
+        Node::IsA { left, right } | Node::IsNA { left, right } => if let Node::Id { .. } = right.node {
```
JSAbrahams, line 163: This logic should be implemented in it's own PR.
This changes the logic from `isa` which was an accidental copy past of `is`.
Now, the right must be an identifier identifying a class.

src/check/constrain/unify/function.rs
```diff
@@ -217,7 +217,7 @@ fn unify_fun_arg(
                 return Err(vec![TypeErr::new(pos, &msg)]);
             }
             EitherOrBoth::Right(_) => {
-                let msg = format!("Function takes only {} arguments", f_args.len());
+                let msg = format!("Function takes only {} {}", f_args.len(), if f_args.len() > 1 { "arguments" } else { "argument" });
```
JSAbrahams, line 220: This can remain here I think.

src/check/resource/primitive/bool.py
```diff
@@ -1,5 +1,5 @@
 class bool:
-    def __init__(self, arg: bool) -> bool: pass
+    def __init__(self, arg: Union[int, str, bool]) -> bool: pass
```
JSAbrahams, line 2: Should be in it's own PR where we update the constructor of primitives.

src/check/resource/primitive/float.py
```diff
@@ -1,7 +1,7 @@
 from typing import Union
 
 class float(complex):
-    def __init__(self, arg: float) -> float: pass
+    def __init__(self, arg: Union[str, float]) -> float: pass
```
JSAbrahams, line 4: Should be in it's own PR where we update the constructor of primitives.

src/check/resource/primitive/int.py
```diff
@@ -1,7 +1,7 @@
 from typing import Union
 
 class int(float):
-    def __init__(self, arg: int) -> int: pass
+    def __init__(self, arg: Union[str, int]) -> int: pass
```
JSAbrahams, line 4: Should be in it's own PR where we update the constructor of primitives.

src/check/resource/primitive/string.py
```diff
@@ -10,6 +10,8 @@ def __eq__(self, other: str) -> bool: pass
 
     def __iter__(self) -> str_iterator: pass
 
+    def is_digit(self) -> bool: pass
```
JSAbrahams, line 13: Should be in it's own PR where we update primitives.

src/check/resource/std/input
```diff
@@ -0,0 +1 @@
+def input(in: str) -> str: pass
```
JSAbrahams, line 1: Should be in it's own PR where we update the standard library.

tests/resource/valid/operation/boolean.mamba
```diff
@@ -17,5 +17,5 @@ def j := False
 
 c is d
 e isnt f
-g isa h
-i isnta j
+g isa Exception
```
JSAbrahams, line 20: Belongs in PR where we fixed `isa` logic

tests/resource/valid/operation/boolean_check.py
```diff
@@ -17,5 +17,5 @@
 
 c is d
 e is not f
-isinstance(g, h)
-not isinstance(i, j)
\ No newline at end of file
+isinstance(g, Exception)
```
JSAbrahams, line 20: Belongs in PR where we fixed `isa` logic

JSAbrahams (commented):
All these changes should again be moved to their own PR's (and get their own tests)

PR #390, opened 2022-12-27
In define mode in handle condition

JSAbrahams:
### Summary

- Don't allow return outside function
- Preserve variable defined in handle cond
 
### Added Tests

*Happy*
- Handle based on README example

*Sad*
- Empty return outside function
- Return outside function


tests/common.rs
```diff
@@ -61,7 +60,7 @@ pub fn resource_content_randomize(valid: bool, subdirs: &[&str], file: &str) ->
 
 #[cfg(test)]
 pub fn resource_content(valid: bool, subdirs: &[&str], file: &str) -> String {
-    resource_content_path(&resource_path(valid, subdirs, file))
+    resource_content_path(&resource_path(valid, subdirs, file)).unwrap()
```
JSAbrahams, line 63: Bit unfortunate, but rewriting all tests which use `resource_content` to deal with a `Result` is a bit much for this PR.

PR #393, opened 2022-12-27
Update std primitives

JSAbrahams:
### Summary

- Allow primitives to take union of types.
- Simplify equality for `Expect`.
- At function call sites, arguments may only be or be child of argument type.

### Added Tests

*Happy*
- `input` std function present.
- all possible inputs for primitives.
- Function call with child of class.

*Sad*
- Function call with super of an argument type.


JSAbrahams:
Doesn't appear to work, when we have:
```
def aa := Complex(10, 20)
```
We get:
```
Error: Unifying two types: Expected a {Int, Float}, was a {String, Float, Int}
   1 | def aa := Complex(10, 20)
                         ^^
```

PR #401, opened 2022-12-28
Run test action on publish

JSAbrahams:
Also run test on release, so that the last badge on the main README refers to the correct build.


JSAbrahams:
Getting duplicates now hm.

JSAbrahams:
This may have been a false alarm.

PR #416, opened 2022-12-30
Add global variable mapping and preserve primitives

### LINKED ISSUE #400: `transpile_file_not_src` test flaky on macosx
### Description of Bug

Build occasionally fails due to this test.
Difficult to reproduce each time as it is flaky.

Runner image info:
```
  Image: macos-11
  Version: 20221215.1
```

### LINKED ISSUE #404: Unifying with fun args creates unnecessary union
### Description of Bug

When unifying variables with function arguments, the union is used.
In some situations this makes sense, but ideally the type should be derived from where the variables is declared first.

### How to Reproduce

Input
```
def factorial(x: Int) -> Int => match x
    0 => 1
    n => n * factorial(n - 1)

def num := input("Compute factorial: ")
if num.is_digit() then
    def result := factorial(Int(num))
    print("Factorial {num} is: {result}.")
else
    print("Input was not an integer.")
```
Creates
```python
from typing import Union
def factorial(x: int) -> int:
    match x:
        case 0:
            return 1
        case n:
            return n * factorial(n - 1)

num: Union[flot, int, str] = input("Compute factorial: ")  # <- here

if num.is_digit():
    result = factorial(int(num))
    print(f"Factorial {num} is: {result}.")
else:
    print("Input was not an integer.")
```

### Expected behavior

Should create
```python
from typing import Union
def factorial(x: int) -> int:
    match x:
        case 0:
            return 1
        case n:
            return n * factorial(n - 1)

num: str = input("Compute factorial: ")  # <- here

if num.is_digit():
    result = factorial(int(num))
    print(f"Factorial {num} is: {result}.")
else:
    print("Input was not an integer.")
```

### LINKED ISSUE #426: Calls to function within current environment not constrained
### How to Reproduce

```
def fun_a() => print("hello world")
def fun_b(x: Int) => print("hello {x}")

fun_a()
fun_b("asdf")
```

### Expected behavior

Should give an error because `"asdf"` is a string but not an int, but no error given.
Surprised we didn't have a sad test for this.

JSAbrahams:
## Relevant issues

- Resolves #404 
- Resolves #426 
- Might have resolved #400 
  We use `output` instead of `ok` to check the output of `Command::main_binary()`
- Made note of #431 
- Made note of #432 

## Summary

Quite a big PR as we had to change some fundamental behaviour.
More fundamentally how we deal with constraint generation.
Diff slightly larger due to some renaming of variables or consistent change in behaviour, namely how we reference `var_mapping`.

### Multiple constraint sets

We now change the behaviour of the `ConstrBuilder` such that we always add constraints to multiple sets in parallel.
This allows us to generate constraints for all possible execution paths (for instance, after an if or match).

We can then either:
- Exit a set locally, thus preventing the constraints here from interacting with constraints further down the line.
- Not exit the set, thus increasing for now the amount of constraints being added to.
  One can use bookmarks (`usize`) so one exits to the correct level at a later stage without knowing how many sets are being added to in parallel.

### Pushing types

Get rid of unnecessary unions by updating logic of how `push_ty` operates.
Whether one has a union of types is determined there and whether any of the `Name`s is interchangeable.

- Streamline further some `Expect` logic which was redundant, which got rid of primitives by directly substituting them.
  Instead, all constraints should be generated in the generate stage, including those for primitives.
  This further simplifies the logic of primitives and makes them more like any other class, which they essentially are (like any good OOP-like language).
- Fix `is_superset_of` logic in `Name`.
  Logic before was erroneous, but not caught due to faulty test.

## Tests

*Happy*
- Modify arithmetic test to get rid of unnecessary union.
- Modify factorial test to get rid of unnecessary union of `input` function output.
*Sad*
- Is superset of Name where other is interchangeable but none of self is super of other
- Like previous, but with same class but other nullable (so self is not super)


JSAbrahams:
Before merging need to document why this works, because the underlying mechanics are starting to become a bit unclear.

JSAbrahams:
Something went wrong after rebase on develop.
So turns out that what we were doing here was actually quite dangerous, could've known.

src/generate/convert/definition.rs
```diff
@@ -42,7 +42,12 @@ pub fn convert_def(ast: &ASTTy, imp: &mut Imports, state: &State, ctx: &Context)
                         match (ty, expression) {
                             (Some(ty), _) => Some(Box::from(convert_node(ty, imp, &state, ctx)?)),
                             (_, Some(expr)) if state.annotate => {
-                                expr.clone().ty.map(|name| name.to_py(imp)).map(Box::from)
+                                let ty = expr.clone().ty.map(|name| name.to_py(imp)).map(Box::from);
+                                if let Some(core) = ty {
```
JSAbrahams, line 46: Clippy warning aside, this logic ideally shouldn't be necessary.
Why do we even get `Core::Empty` in the first place?

JSAbrahams:
It seems that we are now also accidentally allowing any argument, regardless of type, to be passed to functions.
Possibly something wrong with how the any boolean in `Name` affects `is_superset_of`

src/check/constrain/generate/control_flow.rs
```diff
@@ -49,20 +49,24 @@ pub fn gen_flow(
...
-                let then = Expected::try_from((then, &then_env.var_mappings))?;
                 let el = Expected::try_from((el, &else_env.var_mappings))?;
+                constr.add("else branch equal to if", &el, &if_expr_exp);
+            }
+            constr.exit_set(el.pos)?;
 
-                constr.add("if then branch", &if_expr, &then);
-                constr.add("if else branch", &if_expr, &el);
+            if env.is_expr {
+                generate(then, &then_env, ctx, constr)?; // Also gen in outer constr set
```
JSAbrahams, line 68: Shouldn't we also do this for match?
Also add tests

JSAbrahams: We actually already do this for match

JSAbrahams:
Must investigate significant drop in coverage first.

JSAbrahams:
Streamlined quite some logic.
After this, implementing dictionary should hopefully be more straightforward.

tests/resource/valid/class/compound_field_check.py
```diff
@@ -0,0 +1,5 @@
+class MyClass:
+    a: int = 10
+    b: int = self.a + 1
```
JSAbrahams, line 3: Actually, `self` is not defined at the class level, only in methods.

PR #418, opened 2022-12-31
Implement set and list comprehension

### LINKED ISSUE #415: Set and List Generation

- [x] Set generation
- [x] List generation

JSAbrahams:
### Relevant issues

- Resolves #415 

### Summary

Implement set and list builder.

Both in generate and check stage.
First condition is assumed to be a set.

### Added Tests

*Happy*
- Set builder.
- List builder.
- Set/list builder with multiple conditions.
- Nested list comprehensions, to check if environment properly propagates out variables.

*Sad*
- Empty conditions in set builder.
- Undefined variables in conditions.
- Conditions which do not all evaluate to a boolean.


src/check/constrain/generate/operation.rs
```diff
@@ -334,7 +324,9 @@ fn impl_bool_op(
         &Expected::new(ast.pos, &ty),
     );
 
-    generate(left, env, ctx, constr)?;
-    generate(right, env, ctx, constr)?;
-    Ok(env.clone())
+    bin_op(left, right, env, ctx, constr)
+}
+
+fn bin_op(left: &AST, right: &AST, env: &Environment, ctx: &Context, constr: &mut ConstrBuilder) -> Constrained {
```
JSAbrahams, line 330: Is this still necessary?
It shouldn't matter whether we are in define more or not here, right?
Variables shouldn't leak over from one side to the other.

Isn't this just a remnant from a different (wrong) solution?
The solution to our problem, it appears, was treating tuples differently by propagating out the environment when in define mode above.

src/check/constrain/generate/collection.rs
```diff
@@ -18,16 +19,45 @@ pub fn gen_coll(
...
+            constr_col(ast, env, constr, None)?;
+            Ok(res)
         }
-        Node::ListBuilder { .. } => {
-            Err(vec![TypeErr::new(ast.pos, "List builders currently not supported")])
+
+        Node::SetBuilder { item, conditions } | Node::ListBuilder { item, conditions } => {
+            let conds_env = generate(item, &env.is_def_mode(true), ctx, constr)?;
+            if let Some(cond) = conditions.first() {
+                generate(cond, &conds_env.is_def_mode(false), ctx, constr)?;
```
JSAbrahams, line 36: What if this is after the if below?
Ideally we want the error message there to be generated first before any potential ones within the condition. 

We must also add a sad test for if the first condition is not an in operation

PR #423, opened 2023-01-01
Allow wildcard, but with type, in handle arms

### LINKED ISSUE #422: Handle should also work with direct class names
### Current Issue

Handle arms always expect an expression type.
However, what if we don't need to have the name of the error?
Do we then not want to just have an identifier identifying the exception class. 
Alternatively we can throw an error if it is not an expression type, thus always forcing an (empty) identifier.  
This is better than failing silently. 

Either
- allow identifiers, to be used as class names. 
- give an error if using an identifier. 

See generation for handle in check.

JSAbrahams:
### Relevant issues

- Resolves #422 

### Summary

If wildcard identified, convert to different Python expression without `as {id}`.
Also add extra error generation to check stage if:
- Handle does consist of case.
- Condition of arm is not expression type.

Also use let else in collection for less indentation.

### Added Tests

*Happy*
- Handle with identifiers


src/check/constrain/generate/control_flow.rs
```diff
@@ -20,19 +19,22 @@ pub fn gen_flow(
...
-                        Node::ExpressionType { ty: Some(ty), .. } => if let Node::Type { .. } = &ty.node {
-                            if let Ok(name) = TrueName::try_from(ty) {
-                                Some(name)
-                            } else { None }
-                        } else { None }
-                        _ => None
+                        Node::ExpressionType { ty: Some(ty), .. } => TrueName::try_from(ty)
+                            .map_err(|errs| errs.first().expect("At least one").clone()),
+                        other => {
+                            let msg = format!("Expected type identifier, was {other}");
```
JSAbrahams, line 28: Unreachable arm because this is already covered in parse stage.
Perhaps we in future should have a special type of node which can be converted to a `Name` without any error.

Issue #168, opened 2019-11-19
Add hints to error messages


JSAbrahams:
### Current Issue
Error messages are currently still a bit cryptic.
While there are multiple reasons for this, but one is the lack of helpful error messages.

### High-level description of the feature
We should have the ability to add hints strings to error messages.
In the type checker, we might need to create a system that changes the hints being displayed depending on the error message generated before, which would require more granularity in the types of TypeError messages such that we can create conditionals and more fine-tuned error messages.

### Description of potential implementation
First, we need to add a system where we can add hints to an error messages.
Likely an array of strings.

In the type checker, we might need to create a system that changes the hints being displayed depending on the error message generated before, which would require more granularity in the types of TypeError messages such that we can create conditionals and more fine-tuned error messages.

Issue #167, opened 2019-11-19
Properly display code in error messages that covers multiple lines


JSAbrahams:
### Current Issue
Currently, error messages only look good when there is a single code line.

### High-level description of the feature
Allow us to print code blocks with arrows pointing to the culprit.

PR #425, opened 2023-01-02
Prettier error

JSAbrahams:
### Relevant issues

- Should make it easier to implement #168 
- Should make it easier to implement #167 

### Summary

For Type error messages, print the cause of an error as well.
We point to the location with which we're trying to unify for:
- Field and function access.
- Unifying two types.

Streamline print error logic so that it is stored in a single location.
Both simpler logic, and less chance of inconsistent error messages.

### DIFF
--- a/src/common/result.rs
+++ b/src/common/result.rs
@@ -1,8 +1,33 @@
-use std::fmt::Display;
-use std::path::PathBuf;
+use std::cmp::max;
+use std::fmt;
+use std::fmt::{Display, Formatter};
+use std::path::{MAIN_SEPARATOR, PathBuf};
+
+use crate::common::position::Position;
+
+pub const OFFSET_WIDTH: usize = 4;
+
+pub const RIGHT_ARROW: &str = "──→";
+pub const HOOK_ARROW: &str = "└─→";
+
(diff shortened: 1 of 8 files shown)


src/common/result.rs
```diff
@@ -19,3 +44,88 @@ pub fn an_or_a<D>(parsing: D) -> &'static str where D: Display {
...
+
+        format_location(f, 0, None, pos, source)
+    } else {
+        write!(f, "{msg}\n {RIGHT_ARROW} {}\n", path.strip_suffix(MAIN_SEPARATOR).unwrap_or(path))
+    }?;
+
+    let mut first = true;
+    for cause in causes {
+        let msg = cause.msg.as_str().clone();
+        if pos.map_or(false, |pos| pos != cause.pos) && first {
```
JSAbrahams, line 73: `first` is quicker to evaluate so should be first

JSAbrahams:
`We expect [...]` seems a bit off. Just say `Expected [...]`

JSAbrahams:
We have enough helper methods now.
We can start using them properly as we implement others functionality and fix bugs.

JSAbrahams:
Ah, didn't realize I hadn't pushed the commit address the `Expected` comment.
Will fix in future.

Issue #392, opened 2022-12-27
Created pseudo Python syntax so that we can define exceptions for python functions


JSAbrahams:
### Current Issue

Currently the type checker doesn't know whether a std function will throw an exception or not.
Might be nice to allow one to generate a soft warning for exceptions we know might happen.

PR #438, opened 2023-01-08
Parsing and generation of dictionaries

### LINKED ISSUE #414: Dictionaries

- [x] Dictionaries
- [x] Dictionary generation

Nice grammar would be to use `=>` arrow:
```
def my_dict := {
    1 => "One",
    2 => "Two",
    3 => "Three"
}
```
Which would be converted to
```
y_dict = {
    1: "One",
    2: "Two",
    3: "Three"
}
```

We also need to add the collection to the language.
Need to double check the names of methods of dictionary in Python as well.

JSAbrahams:
## Relevant issues

- Do not merge before #439 
- Made note of #443 
- Resolves #414 

## Summary

- Add parsing of dictionaries
- Add dictionary type skeleton to context:
  - iterator returning types
  - `keys()` method which returns iterator where `__next__()` has key return type
  - `values()` method which returns iterator where `__next__()` has value return type
- Check of dictionaries

## Added Tests

*Happy*
- Parse dictionary.
- Parse dictionary builder.
- Dictionary with multiple types.

*Sad*
- Dictionary with types other than signature.
- Access dictionary entries with wrong key type.
- Access dictionary and have use value incorrectly (assume it is other type).
- Access dictionary and have use value incorrectly (assume it is key type).
- Index dictionary using slice
- index set


src/check/resource/std/collection.py
```diff
@@ -57,3 +57,17 @@ class tuple_iterator(Generic[T]):
...
+
+class dict(Generic[T, R]):
+    def __init__(self): pass
+
+    def __iter__(self): dictkeyiterator[T]:
+
+    pass
+
+
+class dictkeyiterator(Generic[T]):
```
JSAbrahams, line 70: Bunch of other iterators for dictionary items which might be nice to include.

tests/generate/collection.rs
```diff
@@ -13,7 +13,7 @@ fn core_list() {
 #[test]
 #[ignore]
```
JSAbrahams, line 14: Surely should no longer be ignored

tests/resource/valid/collection/dictionary_return_is_optional_check.py
```diff
@@ -0,0 +1,5 @@
+from typing import Optional
+j = { 10: 200, 30: 5 }
+
+def f(key: int) -> Optional[int]:
+    return j[key] if key in j else None
```
JSAbrahams, line 5: Actually I don't think we should modify the behaviour here.
Instead, we should deal with `KeyError`.
For now, this error is unchecked then in Mamba.
However, as soon as we have pseudo Python syntax (#392 ) we will be able to specify that for instance a dictionary access can give a `KeyError`, and we will be forced to deal with that.
I think it is too cumbersome to start clumsily implementing optionals.

src/check/constrain/generate/call.rs
```diff
@@ -95,22 +95,17 @@ pub fn gen_call(
...
-            let (exp_col1, exp_col2) = Constraint::iterable("index of collection", &exp_col, &temp_type, &helper_ty);
-            constr.add_constr(&exp_col1, env);
-            constr.add_constr(&exp_col2, env);
-
-            let exp_col_ty = Expected::new(ast.pos, &Type { name: temp_type });
-            constr.add("index of collection", &exp_col_ty, &Expected::from(ast), env);
-
-            generate(item, env, ctx, constr)?;
-            Ok(env.clone())
+            let getitem_ast = AST::new(ast.pos, Node::PropertyCall {
```
JSAbrahams, line 98: Isn't there already logic for magic methods?
Maybe we can call that here we `GET_ITEM` as argument.

PR #439, opened 2023-01-09
Rework Collections for `Expect` and `Name`

### LINKED ISSUE #432: Check trips up when we introduce untyped generics in class signature
### Description of Bug

Now that logic has been generalized, new bugs have come to light.
See generics test.

### LINKED ISSUE #437: Constraint generation can be generalized to remove `Expect::Collection` construct
# Summary of current issue

I think this construct is overly specific and adds unnecessary logic.
For a collection, it is relevant that:
- It has an iterator method: `__iter__()`.
- That said iterator method returns an iterator, and we can get the type of the collection then by checking the type of `__next__`.
- The return type of `__next__` then determines the type of the lookup.

For tuples, there is `TrueName` logic to deal with such constructs.
For literal collections (i.e. defined to top level such as `def x := {2,3,4}`), we can have helper methods which constraint this:
- Iterator method returns a class, and we can pass generics to this class to determine the type of the `__next__` method.
  Then we can proceed as above.
- If we don't know the set type at the generation site, we can probably use a temporary variable.

JSAbrahams:
## Relevant issues

- Do not merge before #440 
- Resolves #437 
- Might resolve #432 

## Summary

As has been a running theme for the past few days, this PR changes quite a lot more than originally intended.
However, we had some sloppy code which I figured would be best to clean up now instead of laborously splitting it over many PR's.
For now I see this a one in a set of major refactors, which for now is fine since I'm the only active developer.
Best to just improve the quality of the codebase now and get rid of as many bugs as possible by generalizing the type system as much as possible.
This includes getting rid of as much superfluous logic as possible.

### Main change: `Except::Collection` to `Expect::Call { ... }`

Remove the collection Expect construct.
Instead, we use generics now and substitution of generics to deal with unification of collections.

- Trim `Collection` type in annotated `AST` since that is not a valid type annotation in Python.
- Add recursive substitution of names.
  Useful if we want to substitute temporary names.
- Add unification and substitution of generics within names.
  Useful if we want to substitute generics which are temporary names.

This greatly simplifies a lot of exceptions we had to make for collections.
We also get to make use of type annotation logic now.

### Secondary: Remove `NameVariant`

Instead, we just make use of `Tuple[...]` and `Callable[Tuple[...],[...]` by making helper methods which create these for us on the fly.
Under the hood, it's always `StringName`.
For now, I leave optionality untouched, so we don't make use of `Optional[...]`, but instead have a boolean in `TrueName`.

This makes it much easier to match names directly and simplifies a lot of substitution logic within `Name`.
The changes here don't affect the codebase outside of `name` too much, because most of it was abstracted away.
But it does greatly reduce the chances of introducing a bug.
And it also simplifies the unification stage even further.

To accomodate these changes, we also revamp the generic system:
- The context now also substitutes the generics within parents when performing a class lookup.
- We perform matching when checking if a class if a class of another class.
  Specifically, if the literal of the names are equal (checking the `lit`s in `StringName { lit1, generics: [Name] }`), we check proceed to check for each generic whether self is name of other as long as the generic count is the same.

## Added Tests

TODO: list tests


src/check/constrain/generate/call.rs
```diff
@@ -100,18 +100,16 @@ pub fn gen_call(
...
-            let temp_collection_type = Type { name: Name::from(temp_type.as_str()) };
-
-            let exp_col = Collection { ty: Box::from(Expected::new(ast.pos, &temp_collection_type)) };
-            let exp_col = Expected::new(ast.pos, &exp_col);
-            constr.add("type of indexed collection", &exp_col, &Expected::from(item), &env);
+            let temp_type = constr.temp_var();
+            let exp_col = Expected::collection(item.pos, &Name::from(temp_type.as_str()));
+            constr.add("type of indexed collection", &exp_col, &Expected::from(item), env);
 
             // Must be after above constraint
```
JSAbrahams, line 107: This comment I think no longer applies, but need to verify this.

src/check/constrain/constraint/builder.rs
```diff
@@ -60,6 +61,15 @@ impl ConstrBuilder {
         trace!("Inserted {var} in constraint builder: {var} => {mapped_var}");
     }
 
+    /// Get a name for a temporary type.
+    ///
+    /// Useful for when we don't know what a type should be during the generation stage.
+    /// The unification stage should then identify these.
+    pub fn temp_var(&mut self) -> String {
```
JSAbrahams, line 68: Rename to `temp_ty` perhaps.
In fact, if we always convert the `String` to a `Name`, we can instead return a `Name`.

JSAbrahams:
Add issue that we should only override identifiers with Python type in:
- types
- constructor calls

src/check/name/true_name/mod.rs
```diff
@@ -19,13 +18,17 @@ pub mod python;
...
-    pub variant: NameVariant,
+    pub variant: StringName,
 }
 
 pub trait IsTemp {
     fn is_temp(&self) -> bool;
 }
 
+pub trait MatchTempName {
+    fn temp_map(&self, other: &StringName, mapping: HashMap<Name, Name>, pos: Position) -> TypeResult<HashMap<Name, Name>>;
```
JSAbrahams, line 29: Hm perhaps some of this logic could be re-used for `match_name`?
They are very similar, except for the if statement.
So if you can pass a predicate then they're the same basically.

src/check/name/mod.rs
```diff
@@ -101,18 +115,10 @@ pub fn match_type_direct(
...
-                let mut mapping = HashMap::with_capacity(1);
-                mapping.insert(id.clone().object(pos)?, (*mutable, Name::from(name)));
-                Ok(mapping)
-            } else {
-                let msg = format!("Cannot match {identifier} with a '{name}'");
-                Err(vec![TypeErr::new(pos, &msg)])
-            }
-        }
-        NameVariant::Tuple(elements) => match identifier {
+    if name.is_tuple() {
```
JSAbrahams, line 118: Isn't there a `Result -> Option` method? Then we can use if-let binding!

src/check/context/clss/mod.rs
```diff
@@ -79,13 +81,27 @@ pub trait GetFun<T> {
...
 impl HasParent<&StringName> for Class {
-    fn has_parent(&self, name: &StringName, ctx: &Context, pos: Position) -> TypeResult<bool> {
-        Ok(self.name.name == COLLECTION && self.name.name == name.name
-            || &self.name == name
-            || self
+    fn has_parent(&self, other: &StringName, ctx: &Context, pos: Position) -> TypeResult<bool> {
+        if self.name == *other || other.name.as_str() == ANY {
+            return Ok(true); // parent of self, of course
+
+            // necessary evil, no way to specify variable generics for tuples
```
JSAbrahams, line 88: Comment should be placed in block below

src/check/constrain/unify/finished.rs
```diff
@@ -1,9 +1,12 @@
 use crate::check::ast::pos_name::PosNameMap;
 use crate::check::context::{Context, LookupClass};
+use crate::check::context::clss::COLLECTION;
 use crate::check::name::{Empty, Name, Union};
 use crate::check::result::TypeResult;
 use crate::common::position::Position;
 
+const IGNORED_NAMES: [&str; 1] = [COLLECTION];
```
JSAbrahams, line 8: Add documentation explaining that IGNORED_NAMES are names which Python does not recognize.
This does unfortunately also make `check` a bit less language agnostic.

src/check/constrain/unify/expression/mod.rs
```diff
@@ -1,42 +1,17 @@
...
-                    }
-                    _ => {
-                        let msg = format!("Expected tuple with {} elements, was {}", elements.len(), ast_elements.len());
-                        return Err(vec![TypeErr::new(left.pos, &msg)]);
-                    }
-                }
-            }
-        }
-
         (Expression { .. }, _) => substitute(constraints, right, left, count, total)?,
```
JSAbrahams, line 15: Surely this file is no longer needed?

src/check/ast/node.rs
```diff
@@ -36,9 +36,9 @@ impl From<&Node> for NodeTy {
                 parents: parents.iter().map(ASTTy::from).collect(),
                 body: body.clone().map(ASTTy::from).map(Box::from),
             },
-            Node::Generic { id, isa } => NodeTy::Generic {
+            Node::Generic { id, .. } => NodeTy::Type {
```
JSAbrahams, line 39: Not a fan, actually.
`NodeTy` should just copy `Node` so the `AST` isn't unnecessarily changed here.
If we want generate rules these should be in the `generate` stage/module.

src/check/name/true_name/mod.rs
JSAbrahams: Do this at the end perhaps when we have a build where all the tests pass again.
Will make it much easier to quickly verify the refactor.

src/check/constrain/constraint/mod.rs
```diff
@@ -58,6 +60,31 @@ impl Constraint {
         Self::access(msg, expected, &Name::from(clss::BOOL), &StringName::from(function::TRUTHY))
     }
 
+    /// Create two constraints for collection:
+    ///
+    /// - A constraint which constraint a call to [ITER] with an iterator, which has an unknown type at this point.
+    ///   To over-bridge this, a temporary type name must be given using [helper_ty].
+    /// - A constraint which constraints [NEXT] call to temporary type with given [col_ty]
+    pub fn collection(msg: &str, expected: &Expected, col_ty: &Name, iter_ty: &Name) -> (Constraint, Constraint) {
```
JSAbrahams, line 68: Perhaps rename to `iterable` since that is more representative of the constraints we are generating.
We are checking that this returns something akin to an iterator (we do duck-typing of sorts because we don't check that is actually an iterator we just check that it implements `__next__()`).

src/generate/convert/definition.rs
```diff
@@ -39,11 +39,9 @@ pub fn convert_def(ast: &ASTTy, imp: &mut Imports, state: &State, ctx: &Context)
                     ty: if matches!(var, Core::TupleLiteral { .. }) {
                         None
                     } else {
-                        match (ty, expression) {
-                            (Some(ty), _) => Some(Box::from(convert_node(ty, imp, &state, ctx)?)),
-                            (_, Some(expr)) if state.annotate => {
-                                expr.clone().ty.map(|name| name.to_py(imp)).map(Box::from)
-                            }
+                        match expression {
+                            Some(expr) if state.annotate =>
```
JSAbrahams, line 43: `matches!` from above can be moved here surely

src/common/position.rs
```diff
@@ -83,6 +83,10 @@ impl Position {
         )
     }
 
+    pub fn invisible() -> Self {
```
JSAbrahams, line 86: Perhaps we should get rid of `Default` implementation for invisible, or make this the default.
Confusing to have both.
Conceptually better to not have default since that doesn't make sense, `invisible` is more descriptive.

src/check/context/clss/mod.rs
```diff
@@ -203,13 +219,17 @@ impl TryFrom<(&GenericClass, &HashMap<Name, Name>, Position)> for Class {
...
         let try_field = |field: &GenericField| Field::try_from((field, generics, pos));
         let try_function = |fun: &GenericFunction| Function::try_from((fun, generics, pos));
+        let try_parent = |parent: &GenericParent| TrueName::try_from((parent, generics, pos));
 
         Ok(Class {
             is_py_type: generic.is_py_type,
-            name: generic.name.substitute(generics, pos)?,
+            name: generic.name.substitute(generics, pos).map_err(|err| err
+                .iter()
+                .map(|e| e.append_msg(&format!("in {}", generic.name)))
```
JSAbrahams, line 228: This append seems unnecessary

src/check/context/clss/mod.rs
```diff
@@ -79,13 +81,27 @@ pub trait GetFun<T> {
...
-            || self
+    fn has_parent(&self, other: &StringName, ctx: &Context, pos: Position) -> TypeResult<bool> {
+        if self.name == *other || other.name.as_str() == ANY {
+            return Ok(true);
+        } else if (self.name.name == TUPLE && (other.name == TUPLE || other.name == COLLECTION)) ||
+            (self.name.name == *other.name && self.name.generics.len() == other.generics.len()) {
+
+            // Contender! check generics
+            // Tuple check is necessary evil, no way to specify variable generics for tuples
+            let mut all_super = true;
```
JSAbrahams, line 92: more descriptive name is `all_generics_super`

src/check/constrain/unify/ty.rs
```diff
@@ -1,39 +1,42 @@
...
-                return unify_link(constraints, finished, ctx, total);
+                substitute_ty(right.pos, r_ty, left.pos, l_ty, constr, count, total)?;
+                return unify_link(constr, finished, ctx, total);
             } else if r_ty.is_temporary() {
-                substitute_ty(left.pos, l_ty, right.pos, r_ty, constraints, count, total)?;
-                return unify_link(constraints, finished, ctx, total);
+                substitute_ty(left.pos, l_ty, right.pos, r_ty, constr, count, total)?;
+                return unify_link(constr, finished, ctx, total);
+            } else if l_ty.contains_temp() {
+                for (old, new) in l_ty.temp_map(r_ty, left.pos)? {
```
JSAbrahams, line 31: The last two blocks in this chain I think supersede the top two.
That is, `temp_map` should also return a map if self is temporary.
And, `contains_temp` is also true if self is temporary.

src/check/constrain/generate/definition.rs
```diff
@@ -164,85 +165,85 @@ pub fn id_from_var(
...
+                    if elements.len() == temp_names.len() {
+                        for (i, (expr, ty)) in enumerate(elements.iter().zip(&temp_names)) {
+                            let expr_exp = Expected::from(expr);
+                            let expr_ty = Expected::new(expr.pos, &Type { name: ty.clone() });
+
+                            let msg = format!("tuple literal element {}", i);
+                            constr.add(&msg, &expr_ty, &expr_exp, &env);
+                        }
+                    } else {
+                        let msg = format!("Cannot join tuple literal. Expected {} elements, was {}",
```
JSAbrahams, line 223: `join` is not descriptive to an end user.
Instead, just say `Expected tuple of {} elements [...]`

src/check/constrain/generate/collection.rs
```diff
@@ -1,118 +1,158 @@
...
+
+            let item = deconstruct_builder(item);
+            let set = AST::new(ast.pos, Node::List { elements: vec![item] });
+            gen_col(&set, &builder_env, constr)
+        }
+        _ => Err(vec![TypeErr::new(ast.pos, "Expected collection")]),
+    }
+}
+
+fn deconstruct_builder(ast: &AST) -> AST {
```
JSAbrahams, line 48: `retrieve_nested_builder_item`

src/check/constrain/unify/link.rs
```diff
@@ -28,30 +27,31 @@ pub fn unify_link(constraints: &mut Constraints, finished: &mut Finished, ctx: &
         trace!("{:width$}[{}{}]  {}", pos, unify, msg, constraint, width = 27);
 
         if let Type { name } = &left.expect {
-            finished.push_ty(ctx, right.pos, name)?;
+            if !name.contains_temp() { finished.push_ty(ctx, right.pos, name)?; }
         }
         if let Type { name } = &right.expect {
```
JSAbrahams, line 32: These two if statements should not be necessary in theory.
We should only add to finished in a successful union of two types.

src/check/constrain/generate/definition.rs
```diff
@@ -164,85 +165,85 @@ pub fn id_from_var(
...
-        }
-    } else {
-        let any = Expected::any(var.pos);
-        for (f_mut, f_name) in identifier.fields(var.pos)? {
-            constr.insert_var(&f_name);
-            env = env.insert_var(mutable && f_mut, &f_name, &any, &constr.var_mapping);
-        }
-    };
+    match (ty, expr) {
+        (Some(ty), Some(expr)) => {
```
JSAbrahams, line 172: Have we tested type and expr?
Should the logic below for `Some(expr)` not also be here?
May be good reason for doing so but maybe just have a small comment then so we don't forget, because on the face of it this looks like a bug.

If it should be here, then better to have wildcard for `ty` and just have a small `if let Some(...) = ...`  binding for ty in the body.

src/check/constrain/unify/link.rs
```diff
@@ -28,30 +27,31 @@ pub fn unify_link(constraints: &mut Constraints, finished: &mut Finished, ctx: &
...
-            (Tuple { .. }, _) | (_, Tuple { .. }) => {
-                unify_type(constraint, constraints, finished, ctx, total)
+            (Expression { .. }, _) => {
+                sub(constraints, right, left, count, total)?;
+                unify_link(constraints, finished, ctx, total)
             }
-            (Type { .. }, _) | (_, Type { .. }) => unify_type(constraint, constraints, finished, ctx, total),
-            (Collection { .. }, Collection { .. }) => {
+            (_, Expression { .. }) => {
+                sub(constraints, left, right, count, total)?;
```
JSAbrahams, line 50: either call both `substitute` and `substitute_ty` or `sub` and `sub_ty`

src/check/constrain/unify/ty.rs
JSAbrahams: Hm apparently not.

src/check/constrain/unify/link.rs
JSAbrahams: Remove it instead in ty

PR #442, opened 2023-01-16
Use `Name` in `NodeTy` where possible

JSAbrahams:
## Summary

This means that generate needs less logic to deal with many edge cases of `ASTTy`.
  Instead, when the check stage generates an annotated `AST` (as `ASTTy`), it converts any type annotations either to `Name`, or `StringName` where relevant.
Generate then no longer needs to deal with generation for different types, only the generation of `Name`, whose logic is well-tested.

## Removed Tests

- If expression is no longer completely annotated in output. Instead, only with one branch.
  Something to do with `is_interchangeable` in `Name`, (see if statement in `Finished`),


src/generate/convert/handle.rs
```diff
@@ -41,11 +35,7 @@ pub fn convert_handle(ast: &ASTTy, imp: &mut Imports, state: &State, ctx: &Conte
                         match &cond.node {
                             NodeTy::ExpressionType { expr, ty, .. } => {
                                 let expr = Box::from(convert_node(expr, imp, state, ctx)?);
-                                let Some(ty) = ty else {
-                                    let msg = format!("Must have condition, was {cond:?}");
-                                    return Err(Box::from(UnimplementedErr::new(cond, &msg)));
-                                };
-                                let class = Box::from(convert_node(ty, imp, state, ctx)?);
+                                let class = Box::from(ty.as_ref().map_or(Core::Empty, |ty| ty.to_py(imp)));
```
JSAbrahams, line 38: Explicit panic here instead of `Core::Empty`

src/generate/convert/handle.rs
```diff
@@ -11,19 +12,12 @@ pub fn convert_handle(ast: &ASTTy, imp: &mut Imports, state: &State, ctx: &Conte
...
             let (var, ty) = if let NodeTy::VariableDef { var, ty, .. } = &expr_or_stmt.node {
-                (
-                    Some(Box::from(convert_node(var, imp, state, ctx)?)),
-                    if let Some(ty) = ty {
-                        Some(Box::from(convert_node(ty, imp, state, ctx)?))
-                    } else {
-                        None
-                    },
-                )
+                (Some(Box::from(convert_node(var, imp, state, ctx)?)), ty.as_ref().map(|ty| ty.to_py(imp)).map(Box::from))
```
JSAbrahams, line 15: move `ty` to previous line.

src/generate/convert/class.rs
```diff
@@ -27,42 +27,33 @@ pub fn convert_class(ast: &ASTTy, imp: &mut Imports, state: &State, ctx: &Contex
...
-            } else {
-                Ok(Core::FunctionCall {
-                    function: Box::from(match convert_node(ty, imp, state, ctx)? {
-                        Core::Type { lit, .. } => Core::Id { lit }, // ignore generics
-                        other => other,
-                    }),
-                    args: convert_vec(args, imp, state, ctx)?,
-                })
-            }
+        NodeTy::Parent { ty, args } => if args.is_empty() {
```
JSAbrahams, line 50: Can't this if match guard?

JSAbrahams:
Drop in coverage is mainly due to dropping unnecessary logic.

Issue #444, opened 2023-01-21
Use Python data model in context


JSAbrahams:
### Current Issue

Now that we have more fully implemented generics, we can leverage this logic to more fully implement the datamodel of Python.
This means that when calling methods, we check the loaded Python classes instead of implementing special logic in the type checker.
We then match methods and their types.

### High-level description of the feature

Call the proper magic methods of Python.
We already make use of:
- `__iter__()`
- `__getitem__(self, item)` (#438 )
- `__str__()`

Up until now, this was done on a as-needed basis.
However, a more exhaustive implementation is possible now thanks to the now more comprehensive generic implementation.

### Description of potential implementation

See https://docs.python.org/3/reference/datamodel.html for method and function names.

JSAbrahams:
A general issue working forward, should be split into constituent components since implementing this all at once is a bit much.

PR #447, opened 2023-01-21
Class inherits all parent fields and functions

JSAbrahams:
## Relevant issues

- Made note of #448 
- Work towards #444 

## Summary

Class now inherits all functions and fields from parent.
This means that a concrete class is complete, and we shouldn't have to resort to any lookups of parents when looking for
fields or function because it is self-contained.

- `in` uses magic method `__contains__`
- Defer to just using `__init__` magic method for consistency, and also get rid of `INIT` token

### Bug Fixes

This also fixes a bug where generics of parent fields and functions were not substituted.
Now this should be no longer an issue.
In general, it is good to use methods of internal resources as much as possible to also uncover potential bugs we might not have covered in tests (such as this one).


src/check/constrain/generate/definition.rs
```diff
@@ -91,7 +95,11 @@ pub fn gen_def(
             let unassigned: Vec<String> = body_env
                 .unassigned
                 .iter()
-                .map(|v| format!("Non nullable class variable '{v}' not assigned to in constructor"))
+                .map(|v| if let Some(class) = &env.class {
+                    format!("Non nullable attribute '{v}' of {class} not assigned to in constructor")
```
JSAbrahams, line 99: Should we actually not also check that we call all parent constructors?
Perhaps good to make an issue of this and add this in separate PR.

src/check/resource/std/collection.py
```diff
@@ -70,25 +72,35 @@ def __getitem__(self, item: T) -> R: pass
...
     def __iter__(self) -> dictkeyiterator[T]: pass
 
+    def __contains__(self, item: T) -> bool: pass
+
     def keys(self) -> dict_keys[T]: pass
 
     def values(self) -> dict_values[R]: pass
 
+
 class dict_keys(Generic[T]):
```
JSAbrahams, line 82: should this and `dict_values` also have a `contains` method?

src/parse/ast/node_op.rs
```diff
@@ -24,17 +25,17 @@ impl Display for NodeOp {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), Error> {
         match &self {
             NodeOp::Assign => write!(f, ":="),
-            NodeOp::Add => write!(f, "{}", function::ADD),
-            NodeOp::Sub => write!(f, "{}", function::SUB),
+            NodeOp::Add => write!(f, "{}", ADD),
+            NodeOp::Sub => write!(f, "{}", SUB),
             NodeOp::Sqrt => write!(f, "{}", function::SQRT),
```
JSAbrahams, line 30: why not also import `function::SQRT` so we can just write `SQRT` here?

Issue #96, opened 2019-05-14
Implement Condition / Type refinement features


JSAbrahams:
### Current Issue
This is only parsed, we do not do anything with conditionals.

### Description of the feature
Use conditionals in the type checker.
In the desugar stage, and if statements when these are encountered and errors where relevant.

PR #451, opened 2023-01-23
Check assigned to in if and match branches

### LINKED ISSUE #449: Consider a field to be assigned to if we do so in every possible branch
## How to Reproduce

```
class MyClass
    def my_field: Int

    def __init__(self):
        if False then 
            self.my_field = 10
        else
            self.my_field = 10
```

## Expected behavior

Should check.

## Additional context

This can be quickly fixed by again re-introducing unions for Environments.
However, this should be done strictly for assign checks, not all variables.

Should also take case that we check that this is for _all_ branches.

JSAbrahams:
## Relevant issues

- Resolves #449 

In future we can perhaps use this logic to also de-nullify types (i.e. `Str? -> Str`) as we move towards implementing type refinement
- #96 

## Summary

If we assign to a variable of a class in all possible execution paths of the constructor, we consider it to be assigned to.

`Environment` intersection and union functions only perform operations on the "must be assigned to" set.
All other items in the `Environment` are left untouched.
That is because
- Shadowing logic is partially implemented in the constraint builder
- We don't want full unions of environments because we don't consider a variable to be declared even if it is declared in all branches (even if this is valid in Python).
  We only consider a variable to be declared if it is declared _in scope_.

## Added Tests

**Happy**
- Assing to class var in all branches of match
- Assign to class var in all branches of if

**Sad**
- Definition does not leak out of then branch...
- ...And else branch of if
- Assign to clas var in only then branch...
- ...And else branch of if
- Assigng to class var in only some branches of match

### DIFF
--- a/src/check/constrain/generate/control_flow.rs
+++ b/src/check/constrain/generate/control_flow.rs
@@ -40,8 +40,7 @@ pub fn gen_flow(
             let outer_env = generate(expr_or_stmt, &env.raises_caught(&raises), ctx, constr)?
                 .raises_caught(&raises_before);
 
-            constrain_cases(ast, &None, cases, &outer_env, ctx, constr)?;
-            Ok(outer_env.clone())
+            constrain_cases(ast, &None, cases, &outer_env, ctx, constr)
         }
 
         Node::IfElse { cond, then, el: Some(el) } => {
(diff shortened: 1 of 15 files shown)


src/check/constrain/generate/control_flow.rs
```diff
@@ -50,19 +49,19 @@ pub fn gen_flow(
...
 
             constr.branch("if else branch", el.pos);
-            generate(el, env, ctx, constr)?;
+            let else_env = generate(el, env, ctx, constr)?;
             if env.is_expr {
                 constr.add("else branch equal to if", &if_expr_exp, &Expected::from(el), env);
             }
 
             constr.reset_branches();
-            Ok(env.clone())
```
JSAbrahams, line 65: Don't variables declared in if branches leak out?

src/check/constrain/generate/control_flow.rs
```diff
@@ -50,19 +49,19 @@ pub fn gen_flow(
...
+            Ok(env.intersection(&then_env.union(&else_env)))
```
JSAbrahams, line 64: A comment perhaps to explain why we have an intersection, might not be obvious at first sight.
Or, better still better documentation for both `intersection` and `union` with an example.

JSAbrahams: Actually the documentation for now appears clear.

